PR #3, opened 2019-03-04
Add reference golang client

### LINKED ISSUE #1: Import go client from ndt-server and/or other sources
I've written several of them. Let's consolidate all of that in a good codebase!

bassosimone:
Closes #1


pboothe:
No tests? :(

pboothe:
Also, I would prefer it be in a directory named `/go` instead of `/golang`

pboothe:
The `Client` struct and its code should be put in a different non-`main` package to allow others to import it.

bassosimone:
> No tests? :(

Yep, more precisely:

_No time, no tests
Another race of conditions
The sea of the simulation
Keep your failures in memory
I fail it, especially tonight_

(With apologies [to Franco Battiato](https://www.youtube.com/watch?v=-Y44YzIODw0))

Please check again and let me know if you like what you see. (What server should I use for integration testing? I recall some mlab4s already run ndt7, but I'd like to ask mlab-ns for geo options.)

bassosimone:
(Oh, I don't have enough permissions to add travis and coveralls, otherwise I'd done it myself.)

pboothe:
It is now enabled on travis, so all you have to do is add a `.travis.yml` file.

pboothe:
For your integration tests, I would encourage you to:
* Only run them when `-short` is not passed to `go test`
* Run them inside a docker container, and cause them to connect to a real `ndt-server` instance also running inside that container.
* Cause them to connect to a local fake `mlab-ns` instance that returns `localhost` at the ndt7 URL either by overriding DNS settings or by making the name of the mlab-ns server a commandline parameter.

If you do these three things, then the integration tests will be the same for every client, regardless of implementation language.

pboothe:
If you add a .travis.yml file in the root of the repo, then I think Travis will start testing it.

pboothe:
Please test the error handling cases too. `go test ./... -short -cover=1` should return 100% for the client library, and testing that no lines of `main` crash looks pretty easy as well (you get 100% coverage when it is trying to run in insecure mode and can't connect).

pboothe:
Here is a cool trick I learned:

If you create a file called `client_test.go` and use it to expose a new function `UpdateDeps`, then that new function will only be available to test code.

Then, in `client.go` you can have:
```go
dialerDial := dialer.Dial
```

and in `client_test.go` you would have:
```go
package client

func UpdateDeps(dial func (URL string, header http.Header)(websocketConn, *http.Response, error)) {
  dialerDial = dial
}
```

And then you can use `UpdateDeps` to avoid having to do all that dancing around setting up a new dependencies object, an approach which strikes me as kind of heavyweight.

Alternatively, you could create a subdirectory of `ndt7` called `fakeserver` and then call `fakeserver.Start()` and have your client test against that server. Creating good fakes is the nice thing to do if you are creating a new service.

go/ndt7-client/main.go
```diff
@@ -0,0 +1,29 @@
...
+	"flag"
+
+	"github.com/apex/log"
+	"github.com/apex/log/handlers/cli"
+	"github.com/m-lab/ndt7-clients/go/ndt7-client/client"
+)
+
+var hostname = flag.String("hostname", "localhost", "Host to connect to")
+var port = flag.String("port", "443", "Port to connect to")
+var insecure = flag.Bool("insecure", false, "Skip TLS verify")
```
pboothe: Recommend a positive condition instead of a negative one.

```go
var secure = flag.Bool("secure", true, "Verify TLS connections.")
```

"not-insecure" is a pretty awkward way of saying "secure".

Or even, instead of using the generic term "secure", how about:

```go
var tlsVerify = flag.Bool("verify-tls", true, "Verify all TLS connections using PKI. Set to false to allow self-signed, spoofed, and hacked certificates.")
```

go/ndt7-client/main.go
```diff
@@ -0,0 +1,29 @@
...
+
+func main() {
+	log.SetHandler(cli.Default)
+	log.SetLevel(log.DebugLevel)
+	flag.Parse()
+	var clnt client.Client
```
pboothe: Shorter and more idiomatic is:
```go
clnt := client.Client{
  Hostname: *hostname,
  Port: *port,
  Insecure: *insecure,
}
```

go/ndt7-client/main.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/m-lab/ndt7-clients/go/ndt7-client/client"
+)
+
+var hostname = flag.String("hostname", "localhost", "Host to connect to")
+var port = flag.String("port", "443", "Port to connect to")
+var insecure = flag.Bool("insecure", false, "Skip TLS verify")
+
+func main() {
+	log.SetHandler(cli.Default)
+	log.SetLevel(log.DebugLevel)
```
pboothe: it intuitively seems like these two lines belong in `init()` and not `main()` but I am not sure why.

go/ndt7-client/sink/reader.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+// logmeasurement logs a measurement received from the server. We trim the
+// trailing newlines such that the logs are more compact.
+func logmeasurement(data []byte) {
+	log.Infof("%s", strings.TrimRight(string(data), "\n"))
+}
+
+// Reader reads messages from the websocket connection in a background
+// goroutine. The length of messages will be posted on the returned
```
pboothe: "The length of each message will be"

go/ndt7-client/source/reader.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"strings"
+	"time"
+
+	"github.com/apex/log"
+	"github.com/gorilla/websocket"
+)
+
+// logmeasurement logs a measurement received from the server. We trim the
```
pboothe: The same code appearing twice indicates the need for a measurementlogger interface, an instance of which you pass in to the Reader object. That way, test code can verify the the right stuff is read, and production code can format the message appropriately.

I also allows others who use this code a very convenient spot to put their own implementations.

go/ndt7-client/source/reader.go
```diff
@@ -0,0 +1,48 @@
...
+func logmeasurement(data []byte) {
+	log.Infof("%s", strings.TrimRight(string(data), "\n"))
+}
+
+// Reader runs in a background goroutine and processes all incoming
+// websocket messages. Text messages are logged. The Reader will
+// post any error that may occur on the returned channel. It will
+// then close the channel. It will close the channel without any
+// error being posted in case of normal websocket closure.
+func Reader(conn *websocket.Conn) <-chan error {
```
pboothe: source.Reader and sink.Reader are SO CLOSE to being the same code. Is there a good way of unifying them?

test/ndt7-testing-backend/main.go
```diff
@@ -0,0 +1,78 @@
+package main
+
```
pboothe: Don't do this. Use the real ndt-server libraries and handlers.

go/ndt7-client/client/client.go
```diff
@@ -0,0 +1,85 @@
...
+	return dialer.Dial(URL, header)
+}
+
+// dial creates and configures the websocket connection
+func (cl Client) dial(urlpath string) (*websocket.Conn, error) {
+	var URL url.URL
+	URL.Scheme = "wss"
+	URL.Path = urlpath
+	URL.Host = cl.Hostname + ":" + cl.Port
+	var dialer websocket.Dialer
```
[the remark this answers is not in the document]
pboothe: Typo:
```go
dialer := websocket.Dialer{
  HandshakeTimeout: 3 * time.Second,
}
```

go/ndt7-client/main.go
bassosimone: I disagree with this specific suggestion because:

1. I believe we want the default to be secure, not insecure 

2. and the default in golang is to initialize to false

I really do not want someone integrating this code and not noticing. Hence, I'd like to have a secure by default approach for this variable. (FWIW, this is also was go stdlib does).

Because of the above reasoning, the main code flow is going to be the code flow where we do the secure stuff, and the insecure branch is gonna be the exception, as in:

```go
   // main flow
  if insecure {
    // do insecure stuff; tell your kids not to do this at home
  }
  // rest
```

go/ndt7-client/main.go
bassosimone: Will do, thanks!

go/ndt7-client/main.go
bassosimone: I think they don't. They configure the global `apex/log` logger. This strikes me as policy, not mechanism, and I argue that the policy depends on how you want to use this code. The current for `main` is optimised to be pleasant. Someone using the `client` package in another context will probably want to use the `json` logger. Also, the current verbosity is maximum, but others may want to have other levels of verbosity for other code.

go/ndt7-client/client/client.go
```diff
@@ -0,0 +1,85 @@
...
+
+	// Port is the port to use
+	Port string
+
+	// Insecure controls whether to skip TLS verification
+	Insecure bool
+}
+
+// dial allows to inject failures when running tests
+var dial = func(dialer websocket.Dialer, URL string, header http.Header) (*websocket.Conn, *http.Response, error) {
```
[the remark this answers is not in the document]
bassosimone: Good suggestion; will do, thanks!

go/ndt7-client/client/client.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	log.Debugf("Connecting to: %s", URL.String())
+	headers := http.Header{}
+	headers.Add("Sec-WebSocket-Protocol", "net.measurementlab.ndt.v7")
+	dialer.HandshakeTimeout = 3 * time.Second
+	conn, _, err := dial(dialer, URL.String(), headers)
+	if err != nil {
+		return nil, err
+	}
+	// According to the specification we must be prepared to read messages
```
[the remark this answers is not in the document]
bassosimone: Right, I'll do that.

go/ndt7-client/client/client.go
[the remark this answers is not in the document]
bassosimone: ACK; will do

go/ndt7-client/client/client.go
```diff
@@ -0,0 +1,85 @@
...
+	// that are smaller than the following value.
+	conn.SetReadLimit(1 << 17)
+	return conn, nil
+}
+
+// closeandwarn will warn if closing a closer causes a failure
+func closeandwarn(closer io.Closer, message string) {
+	time.Sleep(1 * time.Second)
```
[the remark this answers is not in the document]
bassosimone: So, this was actually a sign that I was dealing with a race condition. It seems the way in which I was pipelining the goroutines was poised to generate such race condition. This has been fixed now, by reorganising the pipeline, so I think it makes sense to resolve this convo.

go/ndt7-client/sink/reader.go
bassosimone: The code seems to be changed, I'll check and see whether the new code also has this issue.

go/ndt7-client/source/reader.go
bassosimone: Right; I can create a specific package for this kind of code.

go/ndt7-client/source/reader.go
bassosimone: Sadly, no. After I reorganized the code to make the race go away, they diverged. To be more specific, in the downloader pipeline, the reader is the source of the pipeline, while in the uploader pipeline it is in the middle of the pipeline. It _could_ still be possible to further compress the code by writing a more abstract implementation, but I'd say this is to be decided after looking again into the changed code. So I am going to resolve this convo for now.

test/ndt7-testing-backend/main.go
bassosimone: Yeah, that is indeed the right thing (TM) to do. However, to do that, I need to go back to the server side and make sure I also implement the upload path. This means I am going to put this PR in the backburner for some days, and work instead on the server.

go/ndt7-client/main.go
pboothe: SGTM

go/ndt7-client/main.go
pboothe: I retract my comment. :)

bassosimone:
Superseded by #4

bassosimone:
Closes because #4 is really more mature than this PR.

PR #4, opened 2019-05-03
Write initial ndt7 client

### LINKED ISSUE #1: Import go client from ndt-server and/or other sources

bassosimone:
Copied from a private branch of mine. Already quite big so that it seems legit to start a review even though we don't have 100% coverage and other goodies yet. We'll reach 100% coverage maybe already during this review and then proceed with merging to master.

Closes #1


ndt7.go
```diff
@@ -0,0 +1,151 @@
...
+	ctx context.Context, conn websocketx.Conn, ch chan<- spec.Measurement,
+)
+
+// defaultUserAgent is the default user agent used by this client.
+const defaultUserAgent = "ndt7-client-go/0.1.0"
+
+// Client is a ndt7 client.
+type Client struct {
+	// Ctx is the client context. It's set using NewClient.
+	Ctx context.Context
```
pboothe: Same comment as for the other context inside a struct.

cmd/ndt7-client/main.go
```diff
@@ -0,0 +1,154 @@
...
+
+var flagBatch = flag.Bool("batch", false, "emit JSON events on stdout")
+
+var flagHostname = flag.String("hostname", "", "optional ndt7 server hostname")
+
+var flagTimeout = flag.Int64(
+	"timeout", 45, "seconds after which the ndt7 test is aborted",
+)
+
+func downloadUpload(
```
pboothe: This is a bad name. How about `runTest` ?

And shouldn't this be a method attached to the client object?  Then you could call `client.RunDownloadTest(ctx, emitter)`

pboothe:
Unstated in my review: Thanks for the excellent test coverage! I really appreciate it!

cmd/ndt7-client/main.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"context"
+
+	"github.com/apex/log"
+	"github.com/m-lab/ndt7-client-go"
+)
+
+func main() {
+	client := ndt7.NewClient(context.Background())
```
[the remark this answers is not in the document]
bassosimone: Will do!

internal/mockable/mockable.go
```diff
@@ -0,0 +1,110 @@
+// Package mockable implements mocks.
```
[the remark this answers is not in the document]
bassosimone: Done!

internal/upload/upload.go
```diff
@@ -0,0 +1,116 @@
...
+	go upload(ctx, conn, out)
+	return out
+}
+
+// Run runs the upload subtest.
+func Run(ctx context.Context, conn websocketx.Conn, ch chan<- spec.Measurement) {
+	defer close(ch)
+	defer conn.Close()
+	go ignoreIncoming(conn)
+	t0 := time.Now()
```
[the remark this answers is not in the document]
bassosimone: Done!

mlabns/mlabns.go
```diff
@@ -0,0 +1,126 @@
...
+// TODO(bassosimone): when ndt7 is deployed on the whole platform, we can
+// stop using the staging mlabns service and use the production one.
+const baseURL = "https://locate-dot-mlab-staging.appspot.com/"
+
+// NewClient creates a new Client instance with mandatory userAgent, and tool
+// name. For running ndt7, use "ndt7" as the tool name.
+func NewClient(ctx context.Context, tool, userAgent string) *Client {
+	return &Client{
+		BaseURL:      baseURL,
+		Ctx:          ctx,
```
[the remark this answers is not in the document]
bassosimone: Ouch, thanks for letting me know, fixed!

ndt7.go
bassosimone: Fixed!

mlabns/mlabns.go
```diff
@@ -0,0 +1,126 @@
...
+		requestMaker: http.NewRequest,
+		requestor:    http.DefaultClient,
```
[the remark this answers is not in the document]
bassosimone: Since I'm already wrapping the request with a context, I'll implement this by passing a sub-context that has the desired timeout. I think that "mlab-ns hasn't responded" should take into account not only the connection itself, but also receiving the response. So, I'm going for the sub-context road rather than limiting myself to a dialer.

BTW, this was a great comment. I have also implemented a timeout for the websocket handshake, using features that gorilla/websocket already has.

Also, when reasoning on this comment, I did some yak shaving which resulted in:

1. the `ndt7.Client` constructor takes a mandatory user-agent and the documentation says mlab may reject empty user agents in the future (so, we nudge people towards identifying them by the structure of the code and the documentation)

2. the `ndt7.Client` now has a `*mlabns.Client` parameter and `ndt7.NewClient` constructs such `mlabns.Client` with default-good arguments -- this change is enabled by the previous one and allows me to make the code less repetitive

(Will push soon.)

cmd/ndt7-client/main.go
bassosimone: I renamed the function. With respect to attaching this as a method to the client, I'd rather not do that. The public API of client is what I consider a minimal natural Go API.

For example, in the current API it's clear when to stop and when to fail.

If I attach also this method to the client, I basically introduce a secondary API that is based on an event emitter. Now, I am not super happy about having more than one public API because that complicates the job of people understanding the package.

Also, in general, I find it simpler to have an API where errors are reported in a sync way rather than in an async way. But errors are async in this API with an emitter.

Yet, I see what you mean here. The code in main should be refactored to be more pretty so that a future refactoring that may make it public is enabled.

Hence, I'm going to fully write the main-related code in a proper way, but I'm going to make it also internal. I don't want to expose APIs that people may rely on.

(I spent some time thinking about all of this, and I enjoyed this review because it forced me to think about what to expose and how to expose it more, thanks)

bassosimone:
@pboothe I should have addressed all your comments! Please take another look!

cmd/ndt7-client/internal/emitter/interactive_test.go
```diff
@@ -0,0 +1,52 @@
+package emitter
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/m-lab/ndt7-client-go/spec"
+)
+
+// TestInteractiveDownloadError is what happens if you
```
pboothe: You aren't asserting anything

cmd/ndt7-client/main.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/m-lab/ndt7-client-go"
+	"github.com/m-lab/ndt7-client-go/cmd/ndt7-client/internal/emitter"
+	"github.com/m-lab/ndt7-client-go/spec"
+)
+
+var flagBatch = flag.Bool("batch", false, "emit JSON events on stdout")
+
+var flagHostname = flag.String("hostname", "", "optional ndt7 server hostname")
+
+var flagTimeout = flag.Int64(
```
pboothe: Make this a `time.Duration` using `flag.Duration`

internal/upload/upload.go
```diff
@@ -0,0 +1,117 @@
...
+func upload(ctx context.Context, conn websocketx.Conn, out chan<- int64) {
+	defer close(out)
+	wholectx, cancel := context.WithTimeout(ctx, params.UploadTimeout)
+	defer cancel()
+	preparedMessage, err := makePreparedMessage(params.BulkMessageSize)
+	if err != nil {
+		return // I believe this should not happen in practice
+	}
+	var total int64
+	for {
```
pboothe: could also do equivalently:
```go
for wholectx.Err() == nil {
```
and then you wouldn't need the select.

cmd/ndt7-client/internal/emitter/batch.go
```diff
@@ -0,0 +1,94 @@
...
+
+import (
+	"encoding/json"
+	"os"
+
+	"github.com/m-lab/ndt7-client-go/spec"
+)
+
+type Batch struct {
+	osStdoutWrite func([]byte) (int, error)
```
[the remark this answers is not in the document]
bassosimone: I've implemented it using 

```Go
type Batch struct {
  io.Writer
}
```

and the passing a mocked `io.Writer` in tests.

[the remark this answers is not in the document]
bassosimone: But

```Go
type Interactive struct {
  out io.Writer
}
```

was actually what I needed, thanks!

ndt7.go
```diff
@@ -0,0 +1,152 @@
...
+	MLabNSClient *mlabns.Client
+
+	// UserAgent is the user-agent that will be used. It's set by
+	// NewClient; you may want to change this value.
+	UserAgent string
+
+	// connectFn is the function for connecting a specific
+	// websocket cnnection. It's set to its default value by
+	// NewClient, but you may override it.
+	connectFn connectFn
```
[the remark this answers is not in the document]
bassosimone: Done!

internal/upload/upload.go
bassosimone: Thanks!

cmd/ndt7-client/main.go
bassosimone: Beautiful, thanks!

cmd/ndt7-client/internal/emitter/interactive_test.go
bassosimone: Changed the code to have proper unit tests that verify that we emit what we should.

internal/download/download_test.go
```diff
@@ -0,0 +1,102 @@
...
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/gorilla/websocket"
+	"github.com/m-lab/ndt7-client-go/internal/mocks"
+	"github.com/m-lab/ndt7-client-go/spec"
+)
+
+// TestReadText is the case where we read text messages.
```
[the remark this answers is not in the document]
bassosimone: Changed to perform more checks, including that we emit reasonable messages and that we actually have an error when we provoke an error and we don't see spurious messages.

internal/upload/upload_test.go
```diff
@@ -0,0 +1,114 @@
...
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/gorilla/websocket"
+	"github.com/m-lab/ndt7-client-go/internal/mocks"
+	"github.com/m-lab/ndt7-client-go/spec"
+)
+
+// TestNormal is the normal test case
```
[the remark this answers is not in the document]
bassosimone: Changed to perform more checks, including that we emit reasonable messages and that we actually have an error when we provoke an error and we don't see spurious messages.

bassosimone:
Did some progress towards better testing, but I'm not done yet. Here's what's missing:

- [ ] proper unit tests for the batch emitter

- [ ] see whether we can have better tests (including maybe integration tests) for `main`

- [ ] request a new review

README.md
```diff
@@ -1,3 +1,5 @@
-# ndt7-client-golang
+[![GoDoc](https://godoc.org/github.com/m-lab/ndt7-client-go?status.svg)](https://godoc.org/github.com/m-lab/ndt7-client-go) [![Build Status](https://travis-ci.org/m-lab/ndt7-client-go.svg?branch=master)](https://travis-ci.org/m-lab/ndt7-client-go) [![Coverage Status](https://coveralls.io/repos/github/m-lab/ndt7-client-go/badge.svg?branch=master)](https://coveralls.io/github/m-lab/ndt7-client-go?branch=master) [![Go Report Card](https://goreportcard.com/badge/github.com/m-lab/ndt7-client-go)](https://goreportcard.com/report/github.com/m-lab/ndt7-client-go)
 
-Reference ndt7 golang client implementation.
+# ndt7 Go client
+
+Reference ndt7 Go client implementation.
```
bassosimone: After this PR is merged, I'll write a `main()` with specific pointers to godoc.org.

bassosimone: https://github.com/m-lab/ndt7-client-go/issues/6

cmd/ndt7-client/internal/emitter/batch_test.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"errors"
+	"testing"
+
+	"github.com/m-lab/ndt7-client-go/spec"
+)
+
+// TestBatchDownloadError is what happens if you
+// run a download and it results in an error.
+func TestBatchDownloadError(t *testing.T) {
```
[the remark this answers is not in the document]
bassosimone: I implemented this suggestion!

bassosimone:
@pboothe thanks for the previous review! I have implemented all the suggestions and tried to applied the spirit of the requested changes wherever possible. Please, take another look.

bassosimone:
Thanks, merging!

PR #11, opened 2019-05-15
Add go.mod and go.sum

### LINKED ISSUE #10: Add go.mod and go.sum files
This is to support inclusion of this client in MK and OONI.

bassosimone:
Closes #10


bassosimone:
Thanks! I'll understand why the build is now failing in one case out of three and then merge!

PR #21, opened 2019-07-03
API CHANGE: NewClient(clientName, clientVersion)

### LINKED ISSUE #16: ndt7-client should identify itself with the server
We're not sending meta information as part of the query string. We should do that, because this will help more precise data analysis, and also it's trivial to do.

bassosimone:
This commit changes the NewClient API to take two arguments, the
clientName and the clientVersion, instead of a userAgent.

We didn't specify how userAgent was made. In theory we expected it
to be composed of `${clientName}/${clientVersion}` but that was not
enforced in any way. Per #16, we want the client to identify with
the server also using the query string.

This is for consistency with what the JavaScript client does.

While there, also provide information on the OS and the architecture
of the client, again as part of the query string.

Closes #16.

### DIFF
--- a/ndt7.go
+++ b/ndt7.go
@@ -131,9 +131,7 @@ func (c *Client) discoverServer(ctx context.Context) (string, error) {
 }
 
 // doConnect establishes a websocket connection.
-func (c *Client) doConnect(
-	ctx context.Context, URLPath string,
-) (*websocket.Conn, error) {
+func (c *Client) doConnect(ctx context.Context, URLPath string) (*websocket.Conn, error) {
 	URL := url.URL{}
 	URL.Scheme = "wss"
 	URL.Host = c.FQDN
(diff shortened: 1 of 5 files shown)


pboothe:
Please resolve the conflicts?

bassosimone:
Sure!

bassosimone:
@pboothe done!

ndt7.go
```diff
@@ -110,20 +131,32 @@ func (c *Client) discoverServer(ctx context.Context) (string, error) {
 }
 
 // doConnect establishes a websocket connection.
-func (c *Client) doConnect(ctx context.Context, URLPath string) (*websocket.Conn, error) {
+func (c *Client) doConnect(
```
pboothe: This new function declaration syntax: Why?

bassosimone: I wanted to break long lines and invoking `go fmt ./...` did not cause any change in the code base, so I assumed this formatting was really okay.

Issue #23, opened 2019-07-09
upload: process counter-flow messages


bassosimone:
In https://github.com/m-lab/ndt-server/pull/148 I'm rewriting the upload path so that we also get counter-flow messages during the upload. This PR processes them and makes sure we return to the caller both server- and client-generated messages.

bassosimone:
This is a P2 for me now, as we're currently fine with sender side measurements.

robertodauria:
I believe this is now done: https://github.com/m-lab/ndt7-client-go/pull/38

PR #24, opened 2019-07-09
upload: process counter-flow messages

bassosimone:
Closes https://github.com/m-lab/ndt7-client-go/issues/23

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/24)

### DIFF
--- a/internal/merger/merger.go
+++ b/internal/merger/merger.go
@@ -1,0 +1,37 @@
+// Package merger contains code for merging channels.
+package merger
+
+import (
+	"sync"
+
+	"github.com/m-lab/ndt7-client-go/spec"
+)
+
+// Merge merges the content of two channels into the returned channel.
+func Merge(leftch, rightch <-chan spec.Measurement) <-chan spec.Measurement {
+	// Implementation note: the following is the well known golang
+	// pattern for joining channels together
+	outch := make(chan spec.Measurement)
+	var wg sync.WaitGroup
+	wg.Add(2)
+	// leftch; note that it MUST provide a liveness guarantee
+	go func(out chan<- spec.Measurement) {
+		for m := range leftch {
+			out <- m
+		}
+		wg.Done()
+	}(outch)
+	// rightch; note that it MUST provide a liveness guarantee
+	go func(out chan<- spec.Measurement) {
+		for m := range rightch {
+			out <- m
+		}
+		wg.Done()
+	}(outch)
+	// closer; will always terminate because of above liveness guarantees
+	go func() {
+		wg.Wait()
+		close(outch)
+	}()
+	return outch
+}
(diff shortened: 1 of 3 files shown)


internal/merger/merger.go
```diff
@@ -0,0 +1,37 @@
...
+package merger
+
+import (
+	"sync"
+
+	"github.com/m-lab/ndt7-client-go/spec"
+)
+
+// Merge merges the content of two channels into the returned channel.
+func Merge(leftch, rightch <-chan spec.Measurement) <-chan spec.Measurement {
```
pboothe: Please pass in a `ctx context.Context`, and make each `for` loop exit when the context is canceled.  I suggest doing:

```go
 ok := true
 for ok {
  select {
  case m := <- leftch:
    if m == nil {
      ok = false
    } else {
      out <- m
    }
  case <-ctx.Done():
    ok = false
  }
}
```

Then context cancellation will also cause the `Merge` goroutines to exit. Guaranteeing context cancellation is easier than guaranteeing that channels will be closed. Channel closure is actually remarkably subtle - only writers should close channels, never readers, and only one writer. If a channel is never explicitly closed, but both its endpoints pass out of scope, then the channel will be garbage collected and closed.

internal/upload/upload.go
```diff
@@ -99,38 +128,51 @@ func upload(ctx context.Context, conn websocketx.Conn, out chan<- int64) error {
...
+
 // Run runs the upload subtest. It runs until the ctx is expired or the
 // upload timeout is expired. It uses the provided conn. It emits on the
 // provided channel upload measurements. The returned error is mainly
 // useful for making this function have the same API of download.Run, for
 // which it makes more sense to return an error.
 //
 // Note that run closes both ch and conn.
-func Run(ctx context.Context, conn websocketx.Conn, ch chan<- spec.Measurement) error {
+func Run(
```
pboothe: Why does `Run` return error when the error can only be nil?

internal/merger/merger.go
pboothe: UGH. I wrote a much better comment here and it got eaten. I suggest the following code for Merge:

```go
func Merge(ctx context.Context, chs... <-chan spec.Measurement) <-chan spec.Measurement {
        outch := make(chan spec.Measurement)
￼       var wg sync.WaitGroup
￼       wg.Add(len(chs))
        for _, ch := range chs {
                go func(ch <-chan spec.Measurement) {
                         ok := true
                         for ok {
                                 select {
                                 case m := <-ch:
                                         if m == nil {
                                                 ok = false
                                         } else {
                                                 outch <- m
                                         }
                                 case <-ctx.Done():
                                         ok = false
                                 }
                         }
                         wg.Done()

                }(ch)
        }
        go func() {
￼               wg.Wait()
￼               close(outch)
￼       }()
￼       return outch
}

```

bassosimone:
Superseded by #38

PR #29, opened 2019-09-23
Sync implementation with ndt7 spec v0.8.0

bassosimone:
This mainly boils down to updating the names, making sure we
use application level data for interactive printing.

I've reused m-lab/ndt-server data model, for simplicity.

Also, I didn't know I was not required to document testing
functions when I wrote this code. Now I know, so it is
more simple to not have all these comments.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/29)

### DIFF
--- a/cmd/ndt7-client/internal/emitter/batch.go
+++ b/cmd/ndt7-client/internal/emitter/batch.go
@@ -40,39 +40,45 @@ type batchEvent struct {
 }
 
 type batchValue struct {
-	Failure string
-	Server  string
-	Test    string
+	spec.Measurement
+	Failure string `json:",omitempty"`
+	Server  string `json:",omitempty"`
 }
 
 // OnStarting emits the starting event
-func (b Batch) OnStarting(test string) error {
+func (b Batch) OnStarting(test spec.TestKind) error {
 	return b.emitInterface(batchEvent{
 		Key: "starting",
 		Value: batchValue{
-			Test: test,
+			Measurement: spec.Measurement{
+				Test: test,
+			},
 		},
 	})
 }
 
 // OnError emits the error event
-func (b Batch) OnError(test string, err error) error {
+func (b Batch) OnError(test spec.TestKind, err error) error {
 	return b.emitInterface(batchEvent{
 		Key: "error",
 		Value: batchValue{
+			Measurement: spec.Measurement{
+				Test:    test,
+			},
 			Failure: err.Error(),
-			Test:    test,
 		},
 	})
 }
 
 // OnConnected emits the connected event
-func (b Batch) OnConnected(test, fqdn string) error {
+func (b Batch) OnConnected(test spec.TestKind, fqdn string) error {
 	return b.emitInterface(batchEvent{
 		Key: "connected",
 		Value: batchValue{
+			Measurement: spec.Measurement{
+				Test:   test,
+			},
 			Server: fqdn,
-			Test:   test,
 		},
 	})
 }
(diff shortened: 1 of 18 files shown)


cmd/ndt7-client/internal/emitter/batch.go
```diff
@@ -35,44 +35,44 @@ func (b Batch) emitInterface(any interface{}) error {
 }
 
 type batchEvent struct {
-	Key   string      `json:"key"`
-	Value interface{} `json:"value"`
+	Key   string
+	Value interface{}
```
pboothe: should be `batchValue` not `interface{}`, right?

go.mod
```diff
@@ -2,4 +2,8 @@ module github.com/m-lab/ndt7-client-go
 
 go 1.12
 
-require github.com/gorilla/websocket v1.4.0
+require (
+	github.com/gorilla/websocket v1.4.0
+	github.com/m-lab/ndt-server v0.13.2-0.20190920160430-5b7940b72f0d
```
pboothe: What's this extra stuff here? I'd like to be depend on a real tagged version of ndt-server...

cmd/ndt7-client/internal/emitter/batch.go
bassosimone: No, because there we can also put a Measurement.

bassosimone: As in, I changed to `batchValue` and got:

```
$ go build -v ./cmd/ndt7-client/
github.com/m-lab/ndt7-client-go/cmd/ndt7-client/internal/emitter
# github.com/m-lab/ndt7-client-go/cmd/ndt7-client/internal/emitter
cmd/ndt7-client/internal/emitter/batch.go:84:3: cannot use m (type *spec.Measurement) as type batchValue in field value
cmd/ndt7-client/internal/emitter/batch.go:92:3: cannot use m (type *spec.Measurement) as type batchValue in field value
```

bassosimone: In theory we can do this:

```diff
diff --git a/cmd/ndt7-client/internal/emitter/batch.go b/cmd/ndt7-client/internal/emitter/batch.go
index 1f2a256..5cabf0e 100644
--- a/cmd/ndt7-client/internal/emitter/batch.go
+++ b/cmd/ndt7-client/internal/emitter/batch.go
@@ -36,13 +36,14 @@ func (b Batch) emitInterface(any interface{}) error {
 
 type batchEvent struct {
 	Key   string
-	Value interface{}
+	Value batchValue
 }
 
 type batchValue struct {
-	Failure string
-	Server  string
-	Test    string
+	*spec.Measurement
+	Failure string `json:",omitempty"`
+	Server  string `json:",omitempty"`
+	Test    string `json:",omitempty"`
 }
 
 // OnStarting emits the starting event
@@ -81,7 +82,9 @@ func (b Batch) OnConnected(test, fqdn string) error {
 func (b Batch) OnDownloadEvent(m *spec.Measurement) error {
 	return b.emitInterface(batchEvent{
 		Key:   "measurement",
-		Value: m,
+		Value: batchValue{
+			Measurement: m,
+		},
 	})
 }
 
@@ -89,7 +92,9 @@ func (b Batch) OnDownloadEvent(m *spec.Measurement) error {
 func (b Batch) OnUploadEvent(m *spec.Measurement) error {
 	return b.emitInterface(batchEvent{
 		Key:   "measurement",
-		Value: m,
+		Value: batchValue{
+			Measurement: m,
+		},
 	})
 }
 
```

is it worth it?

pboothe: It is worth it.

`interface{}` and casting is a terrible pattern and we should strive to avoid it whenever possible, for the same reason that we avoid passing around `void*` in C/C++ or `Object` in Java.

bassosimone: Right, then I've made it working! Please, take another look!

PR #33, opened 2020-02-05
ndt7-client: optionally use OpenSSL for TLS

bassosimone:
My objective here is to being able to build `ndt7-client` using
OpenSSL as the TLS engine, which helps a use case that @m-lab has
when running ndt7 with armv7 devices.

I have not touched any public API. This functionality is disabled
by default. The optional `-tags ndt7openssl` command line flag allows
one to compile the client to use OpenSSL. (Further flags such as
`CGO_CFLAGS` and `CGO_LDFLAGS` may be your friends.)

At this stage, I believe we don't want to expose this functionality
to users of this Go library. I consider this being a hack to avoid
the performance issues Go has with armv7.

The idea here is basically that we will use this code to run more
tests and then, if this WAIs, decide what we'll do next.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/33)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: go
 go:
-- 1.12.x
+- 1.13.x
 dist: xenial
 
 before_script:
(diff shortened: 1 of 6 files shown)


bassosimone:
Currently this code is very slow, especially in the upload path.

bassosimone:
Yeah, I agree!

PR #38, opened 2020-02-11
upload: process counterflow messages

bassosimone:
While there, also `go get -u -v ./... && go mod tidy`

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/38)


stephen-soltesz (changes requested):
> Reviewed 1 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/38)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @bassosimone)

___
*[internal/upload/upload.go, line 52 at r3](https://reviewable.io/reviews/m-lab/ndt7-client-go/38#-MeXNw-j9Ei0Tyh8-jaU:-MeXNw-j9Ei0Tyh8-jaV:b-yt5tu7) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/00e6d2c3a267cb95a6ff7774880a1f3bfe02f1b3/internal/upload/upload.go#L52)):*
> ```Go
> 		err := conn.SetReadDeadline(time.Now().Add(params.UploadTimeout))
> 		if err != nil {
> 			errCh <- err
> ```

Previously this loop would `return`. Now it continues -- should this be `continue`? Or return the err and then return? or break? Are these errors recoverable?

Whatever the case, similar handling below.
___
*[internal/upload/upload.go, line 135 at r3](https://reviewable.io/reviews/m-lab/ndt7-client-go/38#-MeXQhIq6whzRO-AT-4i:-MeXQhIq6whzRO-AT-4j:b-ywpvwg) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/00e6d2c3a267cb95a6ff7774880a1f3bfe02f1b3/internal/upload/upload.go#L135)):*
> ```Go
> 	ctx, cancel := context.WithTimeout(ctx, params.UploadTimeout)
> 	defer cancel()
> 	errCh := make(chan error)
> ```

Who closes this channel?
___
*[internal/upload/upload.go, line 148 at r3](https://reviewable.io/reviews/m-lab/ndt7-client-go/38#-MeXPnyy3bRCz7e4XaY9:-MeXPnyy3bRCz7e4XaYA:b-gdpshh) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/00e6d2c3a267cb95a6ff7774880a1f3bfe02f1b3/internal/upload/upload.go#L148)):*
> ```Go
> 
> 	err := <-errCh
> 	if err != nil {
> ```

If Run returns on the first error from the `readcounterflow` goroutine, I would expect the goroutine to quit immediately.

robertodauria (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/38)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[internal/upload/upload.go, line 52 at r3](https://reviewable.io/reviews/m-lab/ndt7-client-go/38#-MeXNw-j9Ei0Tyh8-jaU:-MeZy2B9-VkdjpLtYuUc:b-cvl0p1) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/00e6d2c3a267cb95a6ff7774880a1f3bfe02f1b3/internal/upload/upload.go#L52)):*

Previously this loop would `return`. Now it continues -- should this be `continue`? Or return the err and then return? or break? Are these errors recoverable?

Whatever the case, similar handling below.
</blockquote></details>

Good catch! These should never happen under normal operation.
- Setting a ReadDeadline should always be possible
- ReadMessage() shouldn't fail unless the connection was dropped
- The server should never send anything other than TextMessages during the upload
- The JSON from the server should be a valid `spec.Measurement` in every case

I would prefer to let the upload fail, so it's obvious something is wrong (either on the server or something strange between client and server or the connection was dropped).
___
*[internal/upload/upload.go, line 135 at r3](https://reviewable.io/reviews/m-lab/ndt7-client-go/38#-MeXQhIq6whzRO-AT-4i:-Me_-gaZ4-PavgIVh2C8:bh41y61) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/00e6d2c3a267cb95a6ff7774880a1f3bfe02f1b3/internal/upload/upload.go#L135)):*

Who closes this channel?
</blockquote></details>

The reader (this function). I've added a `defer close(errCh)`.
___
*[internal/upload/upload.go, line 148 at r3](https://reviewable.io/reviews/m-lab/ndt7-client-go/38#-MeXPnyy3bRCz7e4XaY9:-MeZyt1cE9_O8kUiDTVU:blq5kvp) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/00e6d2c3a267cb95a6ff7774880a1f3bfe02f1b3/internal/upload/upload.go#L148)):*

If Run returns on the first error from the `readcounterflow` goroutine, I would expect the goroutine to quit immediately. 
</blockquote></details>

`readcounterflow` now can only write to the channel once (an error or nil if successful) before returning. At this point `readcounterflow` has quit already in every case. Is my understanding correct?

stephen-soltesz (approved):
> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/38)* status: :shipit: complete! 1 of 1 approvals obtained

___
*[internal/upload/upload.go, line 148 at r3](https://reviewable.io/reviews/m-lab/ndt7-client-go/38#-MeXPnyy3bRCz7e4XaY9:-Me_9xW6BBHD3eGIaWTl:b-aztd5r) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/00e6d2c3a267cb95a6ff7774880a1f3bfe02f1b3/internal/upload/upload.go#L148)):*

`readcounterflow` now can only write to the channel once (an error or nil if successful) before returning. At this point `readcounterflow` has quit already in every case. Is my understanding correct?
</blockquote></details>

lgtm :+1:

PR #39, opened 2020-02-12
Add summary at the end of the test

robertodauria:
This change adds a summary at the end of the ndt7 test.

To achieve that, the ndt7 Client now saves the results of each test in a `TestData` struct containing:
1. The latest Measurement sent by the server
2. The latest Measurement sent by the client
3. The latest ConnectionInfo sent by the server

These fields are filled while the test is running, since tests can be interrupted at any time and we do not know which Measurement will be the last one.

Here is how the summary looks like, in interactive mode:
```
$ ./ndt7-client
[...]
         Server: ndt-iupui-mlab3-mil02.measurement-lab.org
         Client: 93.188.101.116
        Latency:    19.8 milliseconds
       Download:   351.9 Mbit/s
         Upload:    55.2 Mbit/s
 Retransmission:    7.98 %
```

And in `-batch` mode
```
$ ./ndt7-client -batch
[...]
{
  "Server": "ndt-iupui-mlab2-mil05.measurement-lab.org",
  "Client": "93.188.101.116",
  "Download": {
    "value": 364.0964194856322,
    "unit": "Mbit/s"
  },
  "upload": {
    "value": 50.754748581668935,
    "unit": "Mbit/s"
  },
  "downloadRetrans": {
    "value": 6.206705982788881,
    "unit": "%"
  },
  "rtt": {
    "value": 18.643,
    "unit": "milliseconds"
  }
}
```

Also, to allow for summary-only output, a `-quiet` flag has been introduced to hide everything but the summary and/or errors.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/39)


cmd/ndt7-client/main.go
```diff
@@ -183,10 +186,23 @@ func main() {
 		InsecureSkipVerify: *flagNoVerify,
 	}
 	r.client.FQDN = *flagHostname
+
+	var e emitter.Emitter
 	if *flagBatch {
-		r.emitter = emitter.NewBatch()
+		e = emitter.NewBatch()
```
pboothe: Flag "batch" should be "json"

cmd/ndt7-client/internal/emitter/interactive.go
```diff
@@ -69,3 +71,41 @@ func (i Interactive) OnComplete(test spec.TestKind) error {
 	_, err := fmt.Fprintf(i.out, "\n%s: complete\n", test)
 	return err
 }
+
+// OnSummary handles the summary event.
+func (i Interactive) OnSummary(s *internal.Summary) error {
+	_, err := fmt.Fprintf(i.out, "%15s: %s\n", "Server", s.Server)
```
pboothe: One big Fprintf, please.

cmd/ndt7-client/main.go
```diff
@@ -183,10 +186,23 @@ func main() {
...
 	} else {
-		r.emitter = emitter.NewInteractive()
+		e = emitter.NewInteractive()
```
pboothe: "Interactive" should be "HumanReadable"

cmd/ndt7-client/main.go
pboothe: I take it back. The flag should be `-format=human` or `-format=json`

pboothe: And default to human

cmd/ndt7-client/internal/summary.go
```diff
@@ -0,0 +1,87 @@
...
+	Value float64
+	Unit  string
+}
+
+// Summary is a struct containing the values displayed to the user at
+// the end of an ndt7 test.
+type Summary struct {
+	// FQDN of the server used for this test.
+	Server string
+	Client string
```
bassosimone: Can you document client?

cmd/ndt7-client/internal/summary.go
```diff
@@ -0,0 +1,87 @@
...
+	// Download speed, in Mbit/s. This is measured at the receiver.
+	Download valueUnitPair
```
bassosimone: I tend to find it more readable when there is a newline after each field, if possible

ndt7.go
```diff
@@ -125,9 +139,11 @@ func NewClient(clientName, clientVersion string) *Client {
 		MLabNSClient: mlabns.NewClient(
 			"ndt7", makeUserAgent(clientName, clientVersion),
 		),
-		upload: upload.Run,
-		Scheme: "wss",
+		upload:  upload.Run,
+		Scheme:  "wss",
+		results: results,
 	}
+
```
bassosimone: Can you not add this empty line?

ndt7.go
```diff
@@ -53,6 +53,14 @@ type testFn = func(
 // by NewClient in the Client.Dialer.HandshakeTimeout field.
 const DefaultWebSocketHandshakeTimeout = 7 * time.Second
 
+// TestData contains the latest Measurement send by the server and the client,
+// plus the latest ConnectionInfo sent by the server.
+type TestData struct {
```
bassosimone: I don't quite like this name because I find it too generic. How about LatestMeasurement? (My understanding is that this data structure is that and, after we fix server, we wont need separate ConnectionInfo)

cmd/ndt7-client/main.go
```diff
@@ -97,11 +99,12 @@ var (
 		Options: []string{"wss", "ws"},
 		Value:   "wss",
 	}
-	flagBatch    = flag.Bool("batch", false, "emit JSON events on stdout")
+	flagJSON     = flag.Bool("json", false, "emit JSON events on stdout")
 	flagNoVerify = flag.Bool("no-verify", false, "skip TLS certificate verification")
 	flagHostname = flag.String("hostname", "", "optional ndt7 server hostname")
 	flagTimeout  = flag.Duration(
 		"timeout", defaultTimeout, "time after which the test is aborted")
+	flagQuiet = flag.Bool("quiet", false, "emit summary and errors only")
```
bassosimone: I quite liked the suggestion by @pboothe to have `-format=json|human|brief` because it seems to me this kinda solves the annoying issue we had with composing emitters in libndt. I am also not so happy with breaking existing scripts that may exist, but, after discussing this with @robertodauria, I understand the potential of this `-format=` design and I am fully on board regarding it. However, please, @robertodauria:

1. if you don't change to `-format=json`, there's a comment above mentioning it, can you fix the comment?

2. if you change the command line arguments, can you also change the docstring and, if needed, the README?

ndt7.go
robertodauria: Your understanding is correct: if ndt-server started sending `ConnectionInfo` with every measurement, rather than just on the first one, then we could theoretically remove the ConnectionInfo field here and just keep the latest Measurement.

However, I feel like we should not rely on the ndt-server specific implementation but rather follow the protocol specification, which at the moment goes in the opposite direction:
```
Servers MUST send this message exactly once.
Clients SHOULD cache the first received instance of this message, and discard any subsequently received instance of this message.
```
To clarify: I like the idea of sending ConnectionInfo every time, I just think we should also update the specification.

bassosimone: Yeah, what you say is sensible.

cmd/ndt7-client/main.go
robertodauria: Following a suggestion you gave me in a private chat, I've kept `-batch` as an alias to `-format=json` but marked it as deprecated, so it shouldn't disrupt anyone's workflow. Also, I've updated the documentation to reflect these changes.

I would argue, however, that it's hard to provide any kind of API stability guarantee when the client is relatively new, and people directly using `master` should expect breaking changes, especially at this stage.

Using semantic versioning + advising users to depend on tagged releases would make this a non-issue. Maybe it's something it would be good to discuss with the team?

cmd/ndt7-client/internal/emitter/humanreadable.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+// HumanReadable is a human readable emitter. It emits the events generated
+// by running a ndt7 test as pleasant stdout messages.
+type HumanReadable struct {
+	out io.Writer
+}
+
+// NewHumanReadable returns a new human readable emitter.
+func NewHumanReadable() HumanReadable {
```
bassosimone: (Shouldn't this be Emitter just like you did for JSON?)

cmd/ndt7-client/internal/emitter/quiet.go
```diff
@@ -0,0 +1,56 @@
+package emitter
+
+import (
+	"github.com/m-lab/ndt7-client-go/cmd/ndt7-client/internal"
+	"github.com/m-lab/ndt7-client-go/spec"
+)
+
+// Quiet acts as a filter allowing summary and error messages only, and
+// doesn't perform any formatting.
+// The message is actually emitted by the embedded Emitter.
```
bassosimone: Do I understand correctly that `-quiet -format=json` or `-quiet -format=human`?

cmd/ndt7-client/internal/summary.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// Summary is a struct containing the values displayed to the user at
+// the end of an ndt7 test.
+type Summary struct {
+	// FQDN of the server used for this test.
+	Server string
+
+	// IP address of the client.
+	Client string
```
bassosimone: Would you mind naming this variable `ClientIP`?

cmd/ndt7-client/internal/summary.go
```diff
@@ -0,0 +1,93 @@
...
+	// Upload speed, in Mbit/s. This is measured at the sender.
+	Upload valueUnitPair
+
+	// Retransmission rate. This is based on the TCPInfo values provided
+	// by the server during a download test.
+	DownloadRetrans valueUnitPair
+
+	// Round-trip time of the latest measurement, in milliseconds.
+	// This is provided by the server during a download test.
+	RTT valueUnitPair
```
bassosimone: Could you follow the pattern `RTT is the round-trip time...`, please?

cmd/ndt7-client/internal/summary.go
```diff
@@ -0,0 +1,93 @@
...
+		if dl.Client.AppInfo != nil && dl.Client.AppInfo.ElapsedTime > 0 {
+			elapsed := float64(dl.Client.AppInfo.ElapsedTime) / 1e06
+			s.Download = valueUnitPair{
+				Value: (8.0 * float64(dl.Client.AppInfo.NumBytes)) /
+					elapsed / (1000.0 * 1000.0),
+				Unit: "Mbit/s",
+			}
+		}
+		if dl.Server.TCPInfo != nil {
+			if dl.Server.TCPInfo.BytesSent > 0 {
```
bassosimone: I appreciate that there are checks for the value being not-zero. Optionally we can write tests for such error condition?

cmd/ndt7-client/main.go
```diff
@@ -80,6 +83,8 @@ import (
 	"os"
 	"time"
 
+	"github.com/m-lab/ndt7-client-go/cmd/ndt7-client/internal"
+
```
bassosimone: I believe this import should be grouped along with others and sorted?

cmd/ndt7-client/main.go
bassosimone: Before providing a more comprehensive response, let me say that the issue we're dealing with here is mostly theoretical, since just me and @mattmathis probably use this client.

Speaking from a theoretical point of view, there is clearly a tension here between moving quickly and breaking things that people may be using. When there is a single repository for all the source code, you can move more quickly because you know that you are not breaking expectations. When there are multiple repositories this is more complex and I believe, from the theoretical point of view, that deleting immediately old code is bad.

Speaking of good practices here, I believe we should consider what the [abseil.io team do in terms of compatibility](https://abseil.io/about/compatibility): "We will support language versions, compilers, platforms, and workarounds as needed for 5 years after their replacement is available, when possible". My suggestion in this regard would be to more modestly strive for keeping the previous public interface for three months. In INTP fashion, of course, here I can also argue that three months equal zero months, because people do not read deprecation notices. Trying to find a synthesis between those two points of view, I'd say we either remove the old interface immediately, or we make it such that people notices the deprecation warning.

(You may also notice that my response here is much more comprehensive than the one I gave you in private as I am making an effort towards not using my [system 1 brain](https://en.wikipedia.org/wiki/Thinking,_Fast_and_Slow) when providing feedback. I think specifically the "three months equal zero months" argument is system 2 thinking.)

[...]

> Using semantic versioning + advising users to depend on tagged releases would make this a non-issue.

I disagree that it is a non-issue. Assuming semantic versioning is done correctly, a semantic bump only tells you that you are going to have some kind of trouble. (In this regard, I believe API stability is impossible, and for this reason it probably makes sense to never provide strong API stability, i.e., never release v.1.0.0.)

> Maybe it's something it would be good to discuss with the team?

This discussion is dangerous material, because certainly it has the potential to go down to a rabbit hole. At the same time, it's also a topic that has to do with engineering, and I may end up learning something, so, yes please!

cmd/ndt7-client/main.go
```diff
@@ -183,10 +191,25 @@ func main() {
 		InsecureSkipVerify: *flagNoVerify,
 	}
 	r.client.FQDN = *flagHostname
-	if *flagBatch {
-		r.emitter = emitter.NewBatch()
+
+	var e emitter.Emitter
+
+	// If -batch, force -format=json.
+	if *flagBatch || *flagFormat == "json" {
```
bassosimone: I don't understand what this code does? Are you checking whether `*flagBatch != nil`?

bassosimone: (I think we can use `flagx` to specify beforehand the available values and then `switch` here. This would strike me as the most @m-lab way of solving this kind of problem; feel free to open an issue if you'd rather not do that now?)

cmd/ndt7-client/main.go
```diff
@@ -183,10 +191,25 @@ func main() {
...
+	}
+	r.emitter = e
+
+	code := r.runDownload(ctx) + r.runUpload(ctx)
+	if code != 0 {
+		osExit(code)
+	}
+
+	s := internal.NewSummary(r.client.FQDN, r.client.Results())
+	r.emitter.OnSummary(s)
```
bassosimone: I agree that it's tricky to decide whether to show a summary if there was an error

ndt7.go
```diff
@@ -170,10 +185,32 @@ func (c *Client) start(ctx context.Context, f testFn, p string) (<-chan spec.Mea
 		return nil, err
 	}
 	ch := make(chan spec.Measurement)
-	go f(ctx, conn, ch)
+	go c.collectData(ctx, f, conn, ch)
 	return ch, nil
 }
 
+func (c *Client) collectData(ctx context.Context, f testFn, conn websocketx.Conn, outch chan<- spec.Measurement) {
```
bassosimone: When we were discussing in private this file's implementation, you mentioned that you'd rather see this piece of code written using interfaces rather than in this function-oriented JavaScript-like fashion. Maybe open an issue?

ndt7.go
```diff
@@ -170,10 +185,32 @@ func (c *Client) start(ctx context.Context, f testFn, p string) (<-chan spec.Mea
...
+	go f(ctx, conn, inch)
+
+	for m := range inch {
+		switch m.Origin {
+		case spec.OriginClient:
+			c.results[m.Test].Client = m
+		case spec.OriginServer:
+			// The server only sends ConnectionInfo once at the beginning of
+			// the test, thus if we want to know the client IP and test UUID
+			// we need to store it separately.
```
bassosimone: Edited https://github.com/m-lab/ndt-server/issues/194#issuecomment-539100964 such that we remember to change this code here.

cmd/ndt7-client/internal/emitter/quiet.go
robertodauria: Yes. In both cases, `-quiet` will only print the summary (+any error) in either JSON or human-readable format.

cmd/ndt7-client/internal/summary.go
robertodauria: At that point I would also need to rename `Server` to `ServerFQDN`, for the same reason.

I originally had `ServerFQDN` and `ClientIP` as field names there, which translated to `server` and `client` in the JSON and you advised me to avoid using the `json:"fieldname"` annotation but rather re-use the same field names in the source code.

Did you change your mind, or you're suggesting to have `{"ServerFQDN": ..., "ClientIP": ...}` in the summary JSON, too?

cmd/ndt7-client/main.go
robertodauria: *flagBatch is `-batch`. It cannot be nil as it has a default value of `false`.

If either one of:
1. `-batch` 
2. `-format=json`

have been passed to the program, the output format will be json. The goal is to handle the deprecation of `-batch` gracefully, in case someone is depending on it. Does it answer your question?

I would not change the flag parsing package here as it feels completely out of the scope for this PR.

cmd/ndt7-client/internal/emitter/quiet.go
robertodauria: Yes. In both cases, -quiet will only print the summary (+ any errors) in the specified format.

ndt7.go
robertodauria: I would like to revisit my opinion about this function-oriented approach after writing the same summary for the ndt5-client (which, AFAIK, is more OO-like?). At the moment I don't feel strongly enough about this to suggest writing it differently.

cmd/ndt7-client/internal/emitter/quiet.go
bassosimone: Yeah, so it is nice that it is possible to use the client in this way.

cmd/ndt7-client/internal/summary.go
bassosimone: When I mentioned to avoid using JSON annotations, I said this because we applied a similar change to the ndt-server in the past, so I wanted to help to enforce consistency in this respect. I don't like the fact that in a JSON variables have upper case names, but I understand how the annotation complicates maintainability.

I believe my comment regarding changing the name of the variable was caused by the documentation string, which starts with "IP", which led me to think in terms of the ClientIP, I guess.

Anyway, variable names can be changed at any time and hence it does not matter.

ndt7.go
bassosimone: Sure. I should actually learn to mind my business more. It's natural that the coding style conventions change when the maintainer of the code changes, and this has already happened. I hope you find ndt5 less functional.

bassosimone:
I apologise for stepping into the review, which I was not asked.

pboothe:
This looks good to me with minor questions. I have faith that, upon considering the questions, you will come to a good conclusion one way or the other, so this PR Looks Good To Me.

PR #45, opened 2020-03-27
Added prometheus exporter

wlbr:
Added a new emitter to generate the prometheus metrics format.
Added a http server that will be startet when "-format=prometheus" is used on the command line.
Added dockerfile to be able to easily install it as a container.

That way it is possible to run the tests driven by prometheus/grafana frequently to generate stats of your upstream connection. I run it on my docker capable Synology, but any local server would be useable. 

**I case you choose to accept the pull request you would need to adapt the contact information in the Dockerfile.**

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/45)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "go.inferGopath": false
+}
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+
 FROM golang:alpine as builder
 
 ENV CGO_ENABLED=0
@@ -26,7 +27,7 @@ RUN adduser -D -g '' appuser
 ADD . ${GOPATH}/src/app/
 WORKDIR ${GOPATH}/src/app
 
-RUN go build -a -installsuffix cgo -ldflags="-w -s" -o /go/bin/mlab_exporter
+RUN go build -a -installsuffix cgo -ldflags="-w -s" -o /go/bin/mlab_exporter cmd/ndt7-client/main.go
 
 # --------------------------------------------------------------------------------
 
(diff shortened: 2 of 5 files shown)


robertodauria:
Hello @wlbr, thank you for contributing and apologies for the delayed reply!

As described in the [Developer Resources](https://www.measurementlab.net/develop/) page, M-Lab imposes a rate-limit of 40 tests per day for interactive users, and 4 per day for integrations / automated testing, to prevent overloading the platform.

For automated tests, we also recommend randomizing test scheduling time over each day, to spread the distribution of tests evenly.

You have probably noticed this already by now, but the late limiting works by sending a certain percentage of empty HTTP responses to the client (with a status code of 204). This percentage quickly approaches 100% if the client keeps making more requests, resulting in the client not finding any available test server.

This happens transparently, i.e. there's nothing saying "you are being rate-limited", but if clients are getting an empty server list from our load balancer that's the most likely reason.

How does that affect this PR? The proposed change would make it _extremely_ easy to unintentionally run over a thousand tests per day, since Prometheus metrics are usually polled every minute. Also, they would be very regular as - AFAIK - Prometheus doesn't allow for a randomized delay between one polling cycle and the next. This is a client behavior we generally don't want to encourage.

Please let me know if you have any questions!

wlbr:
Well, you can adjust Prometheus fetch rhythm by using the scraping-interval setting. This is what I'm doing here 
 https://github.com/wlbr/synology-prometheus , where I integrate some of  these things to monitor my own internet connection. It is set to  once per 10 minutes here. I agree, you cannot randomize that interval.

Actually it was quite helpful in finding some problems with my line and talking to my provider. You may take a look here https://wlbr.de/blog/2020/05/05/dsl-debugging.html , if you are interested.

I would assume that the really small number of folks doing such a nautomated monitoring would be completely invisible in your overall traffic. Especially since the integration to google.com. I generates just <150 calls per day. But yes, the base functionality makes it really easy to create calls really fast. But on the other hand I do not think that this would be a problem, as the people doing this will actually block their own connection simply with speed measurements.

But anyway: I would prefer to keep using mlab, but will change my measurements to using fast.com or something eventually.

PR #62, opened 2021-06-02
Detect CPU capabilities and set scheme accordingly

robertodauria:
This PR changes the default value for the `-scheme` flag to be based on the detected CPU features. The scheme is set to `wss` only on architectures that have hardware AES capabilities so that we don't use TLS where it would otherwise impact performances significantly.

This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/62)


bassosimone (approved):
🐳

robertodauria:
@bassosimone I've changed the code to detect the CPU capabilities via `x/sys/cpu` rather than relying on GOARCH. It looks like it's totally possible for an ARMv8 CPU to lack AES support, since cryptographic extensions are licensed separately by ARM.

PTAL again?

bassosimone (approved):
🐳

Issue #59, opened 2021-05-17
ndt7 vs other speedtest tools


ggmartins:
Dear,

thanks for the great work on nd7 client. It's really easy to download and start using the "app". The retransmissions feature is really interesting. We've been playing around with multiple speed test tools at various vantage points. We found one particular care where Ndt7 has been presenting consistently lower results than the other speedtest tools (namely Oookla's and Iperf UDP). Would you be able to help us to understand why there is? Please, let me know if there's any debug information I can provide to help investigate this "issue". In this case, both Nd7 and Ookla run in sequence from the same device and the command runs from a crontab execution. In the picture below Ndt7 reaches 180Mbps  while Ookla goes all way up to 440Mbps. We have been observing, from other instances, that the speed from all the commands tends to converge at a common speed level, generally.  
I's appreciate any help. Thanks.

![image](https://user-images.githubusercontent.com/2147779/118563258-da61f500-b733-11eb-882f-da6112a47e06.png)

ggmartins:
more info on this:
running Ndt7 from multiple "vantage" points inside the same network:

Jetson NX: `~450Mbps` (ndt7 dw) Linux appflow 4.9.140-tegra 1 SMP PREEMPT Fri Oct 16 12:25:00 PDT 2020 aarch64 aarch64 aarch64 GNU/Linux (Ubuntu 20)
Jetson Nano: `~450Mbps` (ndt7 dw) Linux netrics 4.9.201-tegra 1 SMP PREEMPT Fri Feb 19 08:40:32 PST 2021 aarch64 aarch64 aarch64 GNU/Linux (Ubuntu 20)
Raspberry Pi: `~150Mbps` (ndt7 dw) Linux netrics 5.4.0-1035-raspi 38-Ubuntu SMP PREEMPT Tue Apr 20 21:37:03 UTC 2021 aarch64 aarch64 aarch64 GNU/Linux (Ubuntu 20)
Full output here: https://chicago-cdac.github.io/nm-exp-active-netrics/debug/output.ndt7.txt

good to mention that the degraded ndt7 speed (~150Mbps) is also happening on my teammate with a Comcast connection > 300Mbps dw. He runs the same Raspberry Pi model / OS. 

Thanks,

G

bassosimone:
What is the CPU load when running the test? Is `ndt7-client-go` using 100% of the CPU?

ggmartins:
well, certainly less than 100% for individual cores/threads, 

[...]

PR #63, opened 2021-06-04
download: discard binary messages

robertodauria:
This PR changes the download test to avoid allocating memory for incoming binary messages during the download test, by moving from `conn.ReadMessage()` to `conn.NextReader()` and redirecting the output to `ioutil.Discard`.

This allows for a significant reduction in CPU utilization (~25% according to pprof) and consequently improves performance on resource-constrained hardware (such as embedded ARM devices - Raspberry Pi, ODROID, etc.)

Tested on an ARMv7 ODROID device and further confirmed to improve performance on an ARMv8 (arm64) Raspberry Pi at https://github.com/m-lab/ndt7-client-go/issues/59.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/63)


internal/mocks/mocks.go
```diff
@@ -3,20 +3,33 @@ package mocks
...
+	// conn.NextReader
+	MessageByteArray []byte
+
+	// NextReaderMessageType is the type returned by conn.NextReader
+	NextReaderMessageType int
+
+	// NextReaderResult is the result returned by conn.NextReader
+	NextReaderResult error
+
+	// NextReader determines if the Reader's Read() method must fail
```
bassosimone, line 31:
```suggestion
	// NextReaderMustFail determines if the Reader's Read() method must fail
```

bassosimone (approved):
🐳

Issue #58, opened 2021-03-12
Try next available server on doConnection failure


stephen-soltesz:
The Locate API returns multiple available servers for the client to use. The ndt server may always decline to perform a measurement if the server is overloaded, in lame duck mode, or has detected abusive behavior. So, clients should try the next available server returned by the Locate API to provide resilience to any single server failing to run a measurement.

This section should be in a loop, breaking only once all candidate servers have been tried:

https://github.com/m-lab/ndt7-client-go/blob/10bc591e88fea3c55aa81bbb39812047733a2bb5/ndt7.go#L222-L234

PR #67, opened 2022-04-21
Retry on connection failure when multiple servers are available.

robertodauria:
Closes https://github.com/m-lab/ndt7-client-go/issues/58

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/67)

### DIFF
--- a/ndt7.go
+++ b/ndt7.go
@@ -198,6 +198,23 @@ func (c *Client) nextURLFromLocate(ctx context.Context, p string) (string, error
 	return "", ErrNoTargets
 }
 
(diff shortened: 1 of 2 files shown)


bassosimone (approved):
🐳

robertodauria:
It seems top-level files aren't listed under the "Tree" tab, but they are in the "List" tab:

image

I was a bit surprised by that 100% since that's not what vscode tells me when enabling the coverage overlay. I think other packages (or main) are covering the few remaining paths, and the `-coverpkg` flag to `go test` allows a package's tests to cover code from a different package (basically treating the whole library as a single unit). Nice.

PR #69, opened 2022-04-22
example_test.go requires an aliased import to function.

crwilcox:
While looking into this client, I noticed a small error in your example.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/69)

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -5,7 +5,7 @@ import (
 	"log"
 	"time"
 
-	"github.com/m-lab/ndt7-client-go"
+	ndt7 "github.com/m-lab/ndt7-client-go"
 )
 
 // This shows how to run a ndt7 test.


robertodauria (commented):
Hi, thanks for contributing!

This file is written in the form of a [testable example](https://go.dev/blog/examples), but since the output is not predictable (it runs an actual measurement) we can't take advantage of the `// Output: <expected output>` syntax to verify its output. Thus, if you run it with `go test example_test.go` it'll just output:
```
$ go test example_test.go 
ok      command-line-arguments  0.002s [no tests to run]
```

However, I tried moving `example_test.go` into a different go module, renamed the file to `example.go`, changed the package to `main` and the func name to `main` so it's runnable, and `go run example.go` worked without any aliasing.

May I ask how you are trying to run it and the error you are getting?

> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/69#-:-N0LoHze5uPOEXCY7q0P:b4z8ajm)* status: 0 of 1 approvals obtained

Issue #74, opened 2022-05-13
Upload Test Result Inconsistency


kyle-macmillan:
Hi all, 

I wanted to share some results from a controlled experiment I conducted using this client, as well as ask a question about how NDT7 calculates the final speed. I was interested in seeing how close to the true link capacity the `ndt7-client-go` could report under "perfect" network conditions. However, at low uplink capacities ( < 10Mbps), I found that `ndt7-client-go` reported a speed _greater_ than the link capacity. 

## Preliminary Exploration 
I took a look at how `ndt7-client-go` calculates the final speed and found that it uses information in the `APPInfo` messages, which,
according to the [protocol specification document](https://github.com/m-lab/ndt-server/blob/master/spec/ndt7-protocol.md), report a count of all bytes sent or received. However, in the example code provided in the browser version of NDT7, [`ndt7-js`](https://github.com/m-lab/ndt7-js), the final speed is calculated using the `TCPInfo` messages, which provide network statistics from the `TCP_INFO` socket struct. I confirmed that the [online NDT7 test](https://speed.measurementlab.net/#/) calculates the speed using `TCPInfo` messages. 

Having identified this methodological difference, I conducted an in-lab, controlled experiment to see how different the final speed would be depending on whether the `TCPInfo` messages or the `APPInfo` messages were used. 

## Experimental Setup 
The setup is fairly simple. I use two identical System76 Meerkat Model meer6 desktop computers (Intel 11th Gen i5 @2.4Ghz, 16-GB DDR memory, and up to 2.5Gbps throughput) running Ubuntu 20.04,  one to run the [`ndt-server`](https://github.com/m-lab/ndt-server) and one to run the `ndt7-client-go`. The two computers are then connected via Ethernet cable. 

I use the `tc netem` module to set the link capacity.

## Main results 
![Screenshot from 2022-05-13 15-14-04](https://user-images.githubusercontent.com/67476948/168382917-24ea6e97-6fc7-416d-b77c-1d9996e960b5.png)
[...]

PR #75, opened 2022-05-13
Use server-side measurements during upload test

robertodauria:
This change makes the client compute the upload rate via the `TCPInfo` objects periodically sent by the server.

Client-side measurements had been used in the initial implementation since counterflow message processing wasn't implemented yet (see #38) and server-side measurement were not available.

Ideally this should use `AppInfo` to consistently compute application-level rates in both download and upload, but `AppInfo` isn't currently sent by ndt-server. For now, showing a server-side network-level measurement is better than a client-side application-level measurement which will be overestimated on slow links (see #74).

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/75)

### DIFF
--- a/cmd/ndt7-client/main.go
+++ b/cmd/ndt7-client/main.go
@@ -115,9 +115,9 @@ const (
 )
 
 var (
-	ClientName	= "ndt7-client-go-cmd"
-	ClientVersion	= "0.6.1"
-	flagProfile = flag.String("profile", "",
+	ClientName    = "ndt7-client-go-cmd"
+	ClientVersion = "0.6.1"
+	flagProfile   = flag.String("profile", "",
 		"file where to store pprof profile (see https://blog.golang.org/pprof)")
 
 	flagScheme = flagx.Enum{
@@ -273,12 +273,14 @@ func makeSummary(FQDN string, results map[spec.TestKind]*ndt7.LatestMeasurements
 			}
 		}
 	}
-	// Upload comes from the client-side Measurement during the upload test.
+	// The upload rate comes from the receiver (the server). Currently
+	// ndt-server only provides network-level throughput via TCPInfo.
+	// TODO: Use AppInfo for application-level measurements when available.
 	if ul, ok := results[spec.TestUpload]; ok {
-		if ul.Client.AppInfo != nil && ul.Client.AppInfo.ElapsedTime > 0 {
-			elapsed := float64(ul.Client.AppInfo.ElapsedTime) / 1e06
+		if ul.Server.TCPInfo != nil && ul.Server.TCPInfo.BytesReceived > 0 {
+			elapsed := float64(ul.Server.TCPInfo.ElapsedTime) / 1e06
 			s.Upload = emitter.ValueUnitPair{
-				Value: (8.0 * float64(ul.Client.AppInfo.NumBytes)) /
+				Value: (8.0 * float64(ul.Server.TCPInfo.BytesReceived)) /
 					elapsed / (1000.0 * 1000.0),
 				Unit: "Mbit/s",
 			}
--- a/cmd/ndt7-client/internal/emitter/humanreadable.go
+++ b/cmd/ndt7-client/internal/emitter/humanreadable.go
@@ -58,16 +58,29 @@ func (h HumanReadable) onSpeedEvent(m *spec.Measurement) error {
 	// The specification recommends that we show application level
 	// measurements. Let's just do that in interactive mode. To this
 	// end, we ignore any measurement coming from the server.
-	if m.Origin != spec.OriginClient {
-		return nil
+	switch m.Test {
+	case spec.TestDownload:
+		if m.Origin == spec.OriginClient {
+			if m.AppInfo == nil || m.AppInfo.ElapsedTime <= 0 {
+				return errors.New("missing AppInfo or invalid ElapsedTime")
+			}
+			elapsed := float64(m.AppInfo.ElapsedTime)
+			v := 8.0 * float64(m.AppInfo.NumBytes) / elapsed
+			_, err := fmt.Fprintf(h.out, "\rAvg. speed  : %7.1f Mbit/s", v)
+			return err
+		}
+	case spec.TestUpload:
+		if m.Origin == spec.OriginServer {
+			if m.TCPInfo == nil || m.TCPInfo.ElapsedTime <= 0 {
+				return errors.New("missing TCPInfo or invalid ElapsedTime")
+			}
+			elapsed := float64(m.TCPInfo.ElapsedTime)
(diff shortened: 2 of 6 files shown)


cmd/ndt7-client/main.go
```diff
@@ -115,9 +115,9 @@ const (
 )
 
 var (
-	ClientName	= "ndt7-client-go-cmd"
-	ClientVersion	= "0.6.1"
-	flagProfile = flag.String("profile", "",
+	ClientName    = "ndt7-client-go-cmd"
+	ClientVersion = "0.6.1"
```
bassosimone, line 119: I think it may be worth bumping the version number (maybe 0.7.0), document and make a release

cmd/ndt7-client/main.go
```diff
@@ -273,12 +273,14 @@ func makeSummary(FQDN string, results map[spec.TestKind]*ndt7.LatestMeasurements
 			}
 		}
 	}
-	// Upload comes from the client-side Measurement during the upload test.
+	// The upload rate comes from the receiver (the server). Currently
+	// ndt-server only provides network-level throughput via TCPInfo.
+	// TODO: Use AppInfo for application-level measurements when available.
```
bassosimone, line 278: If the spec does not explain in a clear enough way what "when available" means, then I think it may be worth clarifying the spec. The underlying source of concern is that, when you have some L4 connection terminator, you are not really speaking with the server at TCP level. So the server may assume the bandwidth is much higher so that, basically, at the end of the day, you very seldomly receive any useful AppInfo data.

This issue is basically the head-of-line blocking issue with TCP that IIUC should be fixed with QUIC (which, however, may have other issues that we don't know yet).

bassosimone: To further clarify, I just wanted to provide historical content around why the spec says SHOULD but not MUST. I am now wondering whether perhaps we should further explain why it's SHOULD and not MUST _explicitly_.

cmd/ndt7-client/main.go
robertodauria: I agree -- the version bump will be a separate PR once this one is merged.

cmd/ndt7-client/main.go
robertodauria: If there is an L4 terminator, and the server does not get a Measurement object as often as we would like, should the client still include `AppInfo` in its Measurements? I think none of the clients, nor the server, currently does that. 

I don't think there's any harm in including AppInfo with server measurements, either. If the client is able to get some Measurements (and thus, `TCPInfo`), it will also be able to get `AppInfo`. If no counterflow message can go through, the clients cannot display the rate in any case -- unless we keep client-side measurements as a fallback for such a situation.

Summarizing my understanding here -- please let me know if it's wrong:
- During the download measurement, the client's counterflow messages might not get to the server in case there is an L4 terminator on the path. The server can still use its own `TCPInfo` (specifically, the acked bytes which are hard to falsify) to calculate the rate. The client will just use the number of bytes received / elapsed time.
- During the upload measurement, the server's counterflow messages might not get to the client. In this case, the client does not have TCPInfo (nor AppInfo) to show the rate at which it's uploading.
- If there's no L4 terminator on the path, adding `AppInfo` is beneficial in every case, since it allows both client and server to know application-level rates at the receiver.
- If there's an L4 terminator, and measurement messages can't go through, adding `AppInfo` does not make things worse than they are.

Is that correct? If so, I think the AppInfo object should be treated as mandatory, at least for our reference server and clients. I'm not sure if we should change the spec, but it seems to me that since both `TCPInfo` and `AppInfo` are _optional_ that would mean that a server can potentially not give _any_ feedback to the client during the upload by design, and that feels wrong.

robertodauria:
I have updated all the unit tests and likely fixed a few in the process (`t.Fatalf` cannot be called from a goroutine), the build completed successfully.

cmd/ndt7-client/main.go
bassosimone: We discussed this further in private with @robertodauria and concluded that the spec SHOULD (<- no pun intended) be modified to say that "the server MUST send AppInfo during the upload [...]" and further explain why this information is important in the interest of accurate app-level data collection at the sender. Because the sender (i.e., the client) can always piggy-back ACKs for the data from the server to the client, I see much less of a head of line blocking hazard for counter-flow messages so, IIUC, they should also arrive ~timely.

stephen-soltesz (approved):
:lgtm: - with two planned TODOs for updates to the version and spec.

> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/75#-:-N2IdFsu3q1P6Ib3yeku:bd929xy)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @bassosimone)

___
*[`cmd/ndt7-client/main.go` line 278 at r1](https://reviewable.io/reviews/m-lab/ndt7-client-go/75#-N1zkqV3Yf9KmzUcO2oT-r1-278:-N2IdCnFCR-kCaALeyY5:bdx0diy) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/2bf7cb24493b31504404c41b1d78ea3204400954/cmd/ndt7-client/main.go#L278)):*
<details><summary><i>Previously, bassosimone (Simone Basso) wrote…</i></summary><blockquote>

We discussed this further in private with @robertodauria and concluded that the spec SHOULD (<- no pun intended) be modified to say that "the server MUST send AppInfo during the upload [...]" and further explain why this information is important in the interest of accurate app-level data collection at the sender. Because the sender (i.e., the client) can always piggy-back ACKs for the data from the server to the client, I see much less of a head of line blocking hazard for counter-flow messages so, IIUC, they should also arrive ~timely.
</blockquote></details>

Is this also planned for a future PR?

robertodauria (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/75)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @bassosimone)

___
*[`cmd/ndt7-client/main.go` line 278 at r1](https://reviewable.io/reviews/m-lab/ndt7-client-go/75#-N1zkqV3Yf9KmzUcO2oT-r1-278:-N2IhiJ90y3N2Rk8r_6q:b-8i6mbb) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/2bf7cb24493b31504404c41b1d78ea3204400954/cmd/ndt7-client/main.go#L278)):*

Is this also planned for a future PR?
</blockquote></details>

It's documented on an issue (https://github.com/m-lab/ndt-server/issues/358) -- implementing this change likely means two PRs, one to modify the spec and one to actually add `AppInfo` to the object sent by ndt-server.

Once we have `AppInfo`, I plan on changing the client(s) again to use it.

PR #77, opened 2022-06-11
Add a reference implementation of non-interactive period test runner and export results in a Prometheus-compatible format

fhltang:
Take the algorithm from https://github.com/m-lab/ndt-server/blob/master/spec/ndt7-protocol.md#requirements-for-non-interactive-clients and add it to the ndt7-client command. Specifically, schedule tests using a Poisson point process by selecting inter-test periods from the exponential distribution with an average of 6h (and clipped within the range [36 mins, 15 hours]).

Also export test results using the Prometheus client. This achieves some of the same goals as pull request #45 with the main difference of internal scheduling (using the Poisson point process described above) rather than external scheduling (e.g. Prometheus). The internal scheduler implementation simplifies meeting the "Requirements for non-interactive clients".

Include a Dockerfile to simplify running the ndt7-client command.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/77)


robertodauria (changes requested):
Hi @fhltang, thank you for contributing! Please see my comments below.

> Reviewed 3 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-:-N555RQ98hQv_bryvi4-:b-xz1pr1)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @fhltang)

___
*[`.gitignore` line 4 at r1](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-N559BSW5T5LkTRXW0J-:-N559BSW5T5LkTRXW0J0:b-97xynp) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/cd5dc2689a1073d8ae6fd0a33eefff1a841b805d/.gitignore#L4)):*
> ```ignore
> *~
> [#]*[#]
> [.][#]*
> ```

Which files is this going to ignore?
___
*[`Dockerfile` line 1 at r1](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-N55BOlVDo4B8DEuFR6Q:-N55BOlVDo4B8DEuFR6R:b-8wi2gc) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/cd5dc2689a1073d8ae6fd0a33eefff1a841b805d/Dockerfile#L1)):*
> ```Dockerfile
> FROM golang:1.18-alpine as ndt7-client-build
> ```

Following up to my comment below about moving the Prometheus exporter code into a separate main, this Dockerfile would then run the `ndt7-prometheus-exporter` binary, so we can publish it as `measurementlab/ndt7-prometheus-exporter` on DockerHub. A "standard CLI" Dockerfile to run the current ndt7-client would also be a welcome addition.

(The more I think about this idea of providing an exporter, the more I like it -- thanks again for your contribution.)
___
*[`cmd/ndt7-client/main.go` line 44 at r1](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-N557n946mqeYAOUWQgn:-N557n95DEhnqcTATGJC:b-ba2jpv) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/cd5dc2689a1073d8ae6fd0a33eefff1a841b805d/cmd/ndt7-client/main.go#L44)):*
> ```Go
> // that can be consumed by Prometheus (http://prometheus.io).
> //
> // The `-daemon` flag runs tests repeatedly with rate limiting. It is intended
> ```

I like the option to export test results as Prometheus metrics a lot, but it feels a bit strange to me that a client can run in "daemon mode". What do you think of moving this functionality into a separate command-line tool (in this repository) and make it a Prometheus exporter only? Something like `cmd/ndt7-prometheus-exporter/main.go` would be very nice and can probably be done by moving the code you have added to this file with minimal changes. It also prevents mixing daemon-specific command-line flags with the flags used by ndt7-client when used as a CLI.
___
*[`cmd/ndt7-client/main.go` line 156 at r1](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-N55Ck8eBcEfIraTOkhJ:-N55Ck8eBcEfIraTOkhK:bczrwyk) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/cd5dc2689a1073d8ae6fd0a33eefff1a841b805d/cmd/ndt7-client/main.go#L156)):*
> ```Go
> 	flagDaemon = flag.Bool("daemon", false, "run tests in a (rate limited) loop")
> 	// The flag values below implement rate limiting at the recommended rate
> 	flagPeriodMean = flag.Int("period_mean", 21600, "mean period (in seconds) between speed tests, when running in daemon mode")
> ```

Please use `flag.Duration` here. It allows for any duration parsable by [`time.ParseDuration`](https://pkg.go.dev/time#ParseDuration) and makes the period specification easier to read:

```
ParseDuration parses a duration string. A duration string is a possibly signed sequence of decimal numbers, each with optional fraction and a unit suffix, such as "300ms", "-1.5h" or "2h45m". Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
```
___
*[`cmd/ndt7-client/main.go` line 414 at r1](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-N55FSB6-UA10jfui_gt:-N55FSB6-UA10jfui_gu:b-l37qbc) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/cd5dc2689a1073d8ae6fd0a33eefff1a841b805d/cmd/ndt7-client/main.go#L414)):*
> ```Go
> 		downloadGauge := prometheus.NewGaugeVec(
> 			prometheus.GaugeOpts{
> 				Name: "mlab_ndt7_download",
> ```

Please use a single word to namespace the metric and a suffix with the unit (https://prometheus.io/docs/practices/naming/), e.g.`ndt7_download_rate_bps`. The unit should also be the base one (bit/s instead of Mbit/s). There's also a recommendation to use bytes instead of bits on the Prometheus documentation but since this is network throughput bit/s seems much more appropriate.
___
*[`cmd/ndt7-client/main.go` line 417 at r1](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-N55Fx9Z2B9TNB4xjH7R:-N55Fx9Z2B9TNB4xjH7S:bphh5f4) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/cd5dc2689a1073d8ae6fd0a33eefff1a841b805d/cmd/ndt7-client/main.go#L417)):*
> ```Go
> 				Help: "m-lab ndt7 download speed in Mbit/s",
> 			},
> 			[]string{"client_ip"})
> ```

`server_ip` (or just `client` and `server` if it includes the ports) would also be useful here. I can imagine a private testing infrastructure with multiple clients and servers, or just someone using the Locate load balancer to get an official M-Lab server, thus (likely) getting a different server every time.
___
*[`cmd/ndt7-client/internal/limiter/limiter.go` line 2 at r1](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-N557gVL1xUPuLBcsDq6:-N557gVL1xUPuLBcsDq7:b-8crnp3) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/cd5dc2689a1073d8ae6fd0a33eefff1a841b805d/cmd/ndt7-client/internal/limiter/limiter.go#L2)):*
> ```Go
> // Package limiter contains the ndt7-client rate limiter.
> package limiter
> ```

A few years ago we created the m-lab/go/memoryless [package](https://github.com/m-lab/go/blob/master/memoryless/memoryless.go) which seems to overlap a lot with this one in its scope. It allows to get a `Ticker` that's compatible with `time.Ticker` and can be used as a drop-in replacement. It's used by -- among the others -- ndt-server and pretty well-tested. Could you please use that one?

robertodauria (changes requested):
Hi @fhltang, thank you for contributing! Please see my comments below.

> Reviewed 3 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-:-N555RQ98hQv_bryvi4-:b-xz1pr1)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @fhltang)

___
*[`.gitignore` line 4 at r1](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-N559BSW5T5LkTRXW0J-:-N559BSW5T5LkTRXW0J0:b-97xynp) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/cd5dc2689a1073d8ae6fd0a33eefff1a841b805d/.gitignore#L4)):*
> ```ignore
> *~
> [#]*[#]
> [.][#]*
> ```

Which files is this going to ignore?
___
*[`Dockerfile` line 1 at r1](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-N55BOlVDo4B8DEuFR6Q:-N55BOlVDo4B8DEuFR6R:b-8wi2gc) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/cd5dc2689a1073d8ae6fd0a33eefff1a841b805d/Dockerfile#L1)):*
> ```Dockerfile
> FROM golang:1.18-alpine as ndt7-client-build
> ```

Following up to my comment below about moving the Prometheus exporter code into a separate main, this Dockerfile would then run the `ndt7-prometheus-exporter` binary, so we can publish it as `measurementlab/ndt7-prometheus-exporter` on DockerHub. A "standard CLI" Dockerfile to run the current ndt7-client would also be a welcome addition.

(The more I think about this idea of providing an exporter, the more I like it -- thanks again for your contribution.)
___
*[`cmd/ndt7-client/main.go` line 44 at r1](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-N557n946mqeYAOUWQgn:-N557n95DEhnqcTATGJC:b-ba2jpv) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/cd5dc2689a1073d8ae6fd0a33eefff1a841b805d/cmd/ndt7-client/main.go#L44)):*
> ```Go
> // that can be consumed by Prometheus (http://prometheus.io).
> //
> // The `-daemon` flag runs tests repeatedly with rate limiting. It is intended
> ```

I like the option to export test results as Prometheus metrics a lot, but it feels a bit strange to me that a client can run in "daemon mode". What do you think of moving this functionality into a separate command-line tool (in this repository) and make it a Prometheus exporter only? Something like `cmd/ndt7-prometheus-exporter/main.go` would be very nice and can probably be done by moving the code you have added to this file with minimal changes. It also prevents mixing daemon-specific command-line flags with the flags used by ndt7-client when used as a CLI.
___
*[`cmd/ndt7-client/main.go` line 156 at r1](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-N55Ck8eBcEfIraTOkhJ:-N55Ck8eBcEfIraTOkhK:bczrwyk) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/cd5dc2689a1073d8ae6fd0a33eefff1a841b805d/cmd/ndt7-client/main.go#L156)):*
> ```Go
> 	flagDaemon = flag.Bool("daemon", false, "run tests in a (rate limited) loop")
> 	// The flag values below implement rate limiting at the recommended rate
> 	flagPeriodMean = flag.Int("period_mean", 21600, "mean period (in seconds) between speed tests, when running in daemon mode")
> ```

Please use `flag.Duration` here. It allows for any duration parsable by [`time.ParseDuration`](https://pkg.go.dev/time#ParseDuration) and makes the period specification easier to read:

```
ParseDuration parses a duration string. A duration string is a possibly signed sequence of decimal numbers, each with optional fraction and a unit suffix, such as "300ms", "-1.5h" or "2h45m". Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
```
___
*[`cmd/ndt7-client/main.go` line 414 at r1](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-N55FSB6-UA10jfui_gt:-N55FSB6-UA10jfui_gu:b-l37qbc) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/cd5dc2689a1073d8ae6fd0a33eefff1a841b805d/cmd/ndt7-client/main.go#L414)):*
> ```Go
> 		downloadGauge := prometheus.NewGaugeVec(
> 			prometheus.GaugeOpts{
> 				Name: "mlab_ndt7_download",
> ```

Please use a single word to namespace the metric and a suffix with the unit (https://prometheus.io/docs/practices/naming/), e.g.`ndt7_download_rate_bps`. The unit should also be the base one (bit/s instead of Mbit/s). There's also a recommendation to use bytes instead of bits on the Prometheus documentation but since this is network throughput bit/s seems much more appropriate.
___
*[`cmd/ndt7-client/main.go` line 417 at r1](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-N55Fx9Z2B9TNB4xjH7R:-N55Fx9Z2B9TNB4xjH7S:bphh5f4) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/cd5dc2689a1073d8ae6fd0a33eefff1a841b805d/cmd/ndt7-client/main.go#L417)):*
> ```Go
> 				Help: "m-lab ndt7 download speed in Mbit/s",
> 			},
> 			[]string{"client_ip"})
> ```

`server_ip` (or just `client` and `server` if it includes the ports) would also be useful here. I can imagine a private testing infrastructure with multiple clients and servers, or just someone using the Locate load balancer to get an official M-Lab server, thus (likely) getting a different server every time.
___
*[`cmd/ndt7-client/internal/limiter/limiter.go` line 2 at r1](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-N557gVL1xUPuLBcsDq6:-N557gVL1xUPuLBcsDq7:b-8crnp3) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/cd5dc2689a1073d8ae6fd0a33eefff1a841b805d/cmd/ndt7-client/internal/limiter/limiter.go#L2)):*
> ```Go
> // Package limiter contains the ndt7-client rate limiter.
> package limiter
> ```

A few years ago we created the m-lab/go/memoryless [package](https://github.com/m-lab/go/blob/master/memoryless/memoryless.go) which seems to overlap a lot with this one in its scope. It allows to get a `Ticker` that's compatible with `time.Ticker` and can be used as a drop-in replacement. It's used by -- among the others -- ndt-server and pretty well-tested. Could you please use that one?

fhltang:
> Hi @fhltang, thank you for contributing! Please see my comments below.

You're welcome and thanks for the review comments.

This is a weekend/evening project for me so it might take me a while to actually make the changes for this PR, but I like your suggestions.

Further comments inlined below. (Please excuse the n00b: github PRs are fairly new to me.)

> _[`.gitignore` line 4 at r1](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-N559BSW5T5LkTRXW0J-:-N559BSW5T5LkTRXW0J0:b-97xynp) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/cd5dc2689a1073d8ae6fd0a33eefff1a841b805d/.gitignore#L4)):_
> 
> > ```gitignore
> > *~
> > [#]*[#]
> > [.][#]*
> > ```
> 
> Which files is this going to ignore?

Emacs creates backup files that look like

    prometheus.go~

(note trailing tilde). It also creates temporary files for files with unsaved changes that look like this

    #prometheus.go#
    .#gitignore

The contents of .gitignore just does some escaping to avoid gitignore comment syntax. I can revert commit 661c290 - emacs is not to everyone's taste.

> _[`Dockerfile` line 1 at r1](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-N55BOlVDo4B8DEuFR6Q:-N55BOlVDo4B8DEuFR6R:b-8wi2gc) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/cd5dc2689a1073d8ae6fd0a33eefff1a841b805d/Dockerfile#L1)):_
> 
> > ```dockerfile
> > FROM golang:1.18-alpine as ndt7-client-build
> > ```
> 
> Following up to my comment below about moving the Prometheus exporter code into a separate main, this Dockerfile would then run the `ndt7-prometheus-exporter` binary, so we can publish it as `measurementlab/ndt7-prometheus-exporter` on DockerHub. A "standard CLI" Dockerfile to run the current ndt7-client would also be a welcome addition.

Sounds good.

> (The more I think about this idea of providing an exporter, the more I like it -- thanks again for your contribution.)

I've attached a couple of screenshots of the kinds of graphs you can get with these metrics. (It's such timeseries graphs that motivate me, personally.)

Seconds Since Last Test (there are actually two graph lines here)...
20220621 Seconds Since Last Test 7d

Download Test Results...
20220621 Download Tests Mbps 7d

> _[`cmd/ndt7-client/main.go` line 44 at r1](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-N557n946mqeYAOUWQgn:-N557n95DEhnqcTATGJC:b-ba2jpv) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/cd5dc2689a1073d8ae6fd0a33eefff1a841b805d/cmd/ndt7-client/main.go#L44)):_
> 
> > ```go
> > // that can be consumed by Prometheus (http://prometheus.io).
> > //
> > // The `-daemon` flag runs tests repeatedly with rate limiting. It is intended
> > ```
> 
> I like the option to export test results as Prometheus metrics a lot, but it feels a bit strange to me that a client can run in "daemon mode". What do you think of moving this functionality into a separate command-line tool (in this repository) and make it a Prometheus exporter only? Something like `cmd/ndt7-prometheus-exporter/main.go` would be very nice and can probably be done by moving the code you have added to this file with minimal changes. It also prevents mixing daemon-specific command-line flags with the flags used by ndt7-client when used as a CLI.

This makes sense, especially since it removes the prometheus client dependency from the single-shot client.

I will try to refactor so that code is shared between `cmd/ndt7-client` and `cmd/ndt7-prometheus-exporter` where it makes sense though I expect there will be some forked code for the plumbing through of argv flags that apply to both. I don't think that is such a bad thing but what is your opinion about maintaining automated test coverage for the forked code? Are you okay with some forked tests?

> _[`cmd/ndt7-client/main.go` line 156 at r1](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-N55Ck8eBcEfIraTOkhJ:-N55Ck8eBcEfIraTOkhK:bczrwyk) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/cd5dc2689a1073d8ae6fd0a33eefff1a841b805d/cmd/ndt7-client/main.go#L156)):_
> 
> > ```go
> > 	flagDaemon = flag.Bool("daemon", false, "run tests in a (rate limited) loop")
> > 	// The flag values below implement rate limiting at the recommended rate
> > 	flagPeriodMean = flag.Int("period_mean", 21600, "mean period (in seconds) between speed tests, when running in daemon mode")
> > ```
> 
> Please use `flag.Duration` here. It allows for any duration parsable by [`time.ParseDuration`](https://pkg.go.dev/time#ParseDuration) and makes the period specification easier to read:

Good idea. I don't code in Golang enough to remember such tricks.

> _[`cmd/ndt7-client/main.go` line 414 at r1](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-N55FSB6-UA10jfui_gt:-N55FSB6-UA10jfui_gu:b-l37qbc) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/cd5dc2689a1073d8ae6fd0a33eefff1a841b805d/cmd/ndt7-client/main.go#L414)):_
> 
> > ```go
> > 		downloadGauge := prometheus.NewGaugeVec(
> > 			prometheus.GaugeOpts{
> > 				Name: "mlab_ndt7_download",
> > ```
> 
> Please use a single word to namespace the metric and a suffix with the unit (https://prometheus.io/docs/practices/naming/), e.g.`ndt7_download_rate_bps`. The unit should also be the base one (bit/s instead of Mbit/s). There's also a recommendation to use bytes instead of bits on the Prometheus documentation but since this is network throughput bit/s seems much more appropriate.

Thanks for digging into this. I will follow the guidance in the prometheus documentation.

> _[`cmd/ndt7-client/main.go` line 417 at r1](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-N55Fx9Z2B9TNB4xjH7R:-N55Fx9Z2B9TNB4xjH7S:bphh5f4) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/cd5dc2689a1073d8ae6fd0a33eefff1a841b805d/cmd/ndt7-client/main.go#L417)):_
> 
> > ```go
> > 				Help: "m-lab ndt7 download speed in Mbit/s",
> > 			},
> > 			[]string{"client_ip"})
> > ```
> 
> `server_ip` (or just `client` and `server` if it includes the ports) would also be useful here. I can imagine a private testing infrastructure with multiple clients and servers, or just someone using the Locate load balancer to get an official M-Lab server, thus (likely) getting a different server every time.

I remember in an earlier iteration, I had added a `server` label. I shied away from that because I was concerned about the potential increase in cardinality (in fact there is a "Caution" note in section https://prometheus.io/docs/practices/naming/#labels). However, at a default of 4 pairs of tests per day, I think cardinality is unlikely to be issue (assuming I address the GC issue mentioned next).

It occurred to me that there is another issue here. When successive test use a different server or client, the old timeseries still exists (with the old value). I haven't implemented garbage collection of old timeseries. This can happen even in the current form of the PR, e.g. whenever the client ip changes.

> _[`cmd/ndt7-client/internal/limiter/limiter.go` line 2 at r1](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-N557gVL1xUPuLBcsDq6:-N557gVL1xUPuLBcsDq7:b-8crnp3) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/cd5dc2689a1073d8ae6fd0a33eefff1a841b805d/cmd/ndt7-client/internal/limiter/limiter.go#L2)):_
> 
> > ```go
> > // Package limiter contains the ndt7-client rate limiter.
> > package limiter
> > ```
> 
> A few years ago we created the m-lab/go/memoryless [package](https://github.com/m-lab/go/blob/master/memoryless/memoryless.go) which seems to overlap a lot with this one in its scope. It allows to get a `Ticker` that's compatible with `time.Ticker` and can be used as a drop-in replacement. It's used by -- among the others -- ndt-server and pretty well-tested. Could you please use that one?

I wasn't aware of that package. I will look into using that instead.

robertodauria:
> 
> This is a weekend/evening project for me so it might take me a while to actually make the changes for this PR, but I like your suggestions.
> 

No problem! :)

> 
> Emacs creates backup files that look like
> [...]
> The contents of .gitignore just does some escaping to avoid gitignore comment syntax. I can revert commit [661c290](https://github.com/m-lab/ndt7-client-go/commit/661c290bdc847598ed0812e322c13099aae08dd8) - emacs is not to everyone's taste.

Thanks, I didn't know about Emacs' backup files. It's totally fine, I was just curious.

> I will try to refactor so that code is shared between `cmd/ndt7-client` and `cmd/ndt7-prometheus-exporter` where it makes sense though I expect there will be some forked code for the plumbing through of argv flags that apply to both. I don't think that is such a bad thing but what is your opinion about maintaining automated test coverage for the forked code? Are you okay with some forked tests?
> 

Ideally each `*.go` file has a corresponding `*_test.go`. I don't think code duplication should be a concern for unit tests in general -- i.e. better to optimize to have each test readable in isolation, rather than factoring out common functions.

As an aside: In terms of test coverage we don't always strive for 100%, higher is better but we accept that there are lines that are tricky to test (especially in a `main`) and the same lines also tend to be "obviously" correct. Anywhere over 80% is fine.

> I remember in an earlier iteration, I had added a `server` label. I shied away from that because I was concerned about the potential increase in cardinality (in fact there is a "Caution" note in section https://prometheus.io/docs/practices/naming/#labels). However, at a default of 4 pairs of tests per day, I think cardinality is unlikely to be issue (assuming I address the GC issue mentioned next).
>
> It occurred to me that there is another issue here. When successive test use a different server or client, the old timeseries still exists (with the old value). I haven't implemented garbage collection of old timeseries. This can happen even in the current form of the PR, e.g. whenever the client ip changes.

Prometheus automatically purges old data (default retention time is 15 days I think?), there's no need to remove timeseries manually. It is also pretty good at keeping memory/disk usage low these days. Thinking about cardinality is a good idea, but we don't have nearly enough servers in any single location for that to be a problem (e.g. if you live around London there are 6 M-Lab "sites", each with 3 production servers, even assuming you eventually hit all of them by random chance it's 18 server IPs -- double that if you have a dual IPv4/v6 stack).

Also, that allows filtering rates by server and seeing differences between network paths, since each M-Lab site has a different transit provider (GTT, Level3, Vodafone, TI Sparkle, etc.). I would actually add the server's hostname if it's available, because it makes this kind of grouping much easier with a [`label_replace`](https://prometheus.io/docs/prometheus/latest/querying/functions/#label_replace) to extract the site from the hostname. :)

fhltang:
@robertodauria 

This PR needs more work but I would wanted to check if you had feedback before committing to making sure the documentation is up to date (and to fixing the Travis CI failure).

I think I've addressed your review comments but there is some wonkiness right now.

To make a separate `main()` for `ndt7-prometheus-exporter`, I made the mistake of putting the code in a subdirectory of `cmd/ndt7-client` (commit 6d557eb). I believe this is why Travis CI believes the build is broken.

I think the right thing to do is to make `ndt7-prometheus-exporter/` a sibling directory of `ndt7-client/` which will require moving `cmd/ndt7-client/internal/` up one level in the directory tree so that `ndt7-prometheus-exporter/` can import those internal packages. I'm not sure if this is sufficient to fix Travis CI since it looks like it assumes there's only one main package. I'm also not so familiar with Go's directory conventions so I'd like to hear your opinions.

With regards to the prometheus metric fields, there was an issue where the test result gauges were keyed by `client_ip` and that meant the value became the last test result per `client_ip` but it wasn't possible to determine which test result was the most recent (without going back in time and looking at older values). Adding a `server` label would have made this more obvious: we'd have the latest test result for each `(client_ip, server)` combination. To solve this (without trying to implement deletion of older gauge metrics), I removed all labels from the test result metrics. Instead, I added a "last successful test" metric with the `client_ip` and `server` labels and this latter metric can be joined against the test result metrics. See the comments in `main.go` in commit e0a7971.

I threw out `poisson.go` and `limiter.go` in favour of `memoryless.Ticker` in commit ad1bd65 as you suggested. (I had also considered a channel-based API before since this will open up possibilities such as making it easier to add a "run test now" HTTP end point.) I like the pattern but one of my tests now has to do some yucky poking into the innards of `memoryless.Ticker` to make it testable.

Commit 75a6079 factors out the `runner struct` into its own package so that it can be shared by `ndt7-prometheus-exporter`. Commit 6d557eb creates `ndt7-prometheus-exporter` by forking `ndt7-client` and ripping out some functionality to make it simpler.

Francis

robertodauria (changes requested):
Hi, sorry about the delay, I've been quite busy lately.

You're right about the directory structure. Folders under `cmd/` should match the executable's name, so I believe what we want here is:
- cmd/ndt7-client/main.go
- cmd/ndt7-prometheus-exporter/main.go
- internal/...

There is an unofficial directory structure standard at https://github.com/golang-standards/project-layout#cmd which we try to follow. Regarding `internal`, it would probably be right to use the `internal/` folder at the root of the repository, which both `main.go` (the client's one and the exporter's one) can import and use.

If I'm not mistaken, these are all the metrics currently exported:
```
# HELP ndt7_download_rate_bps m-lab ndt7 download speed in bits/s
# TYPE ndt7_download_rate_bps gauge
ndt7_download_rate_bps 2.1933977314217535e+08
# HELP ndt7_last_success_timestamp_seconds last successful m-lab ndt7 test completion time in seconds since 1970-01-01
# TYPE ndt7_last_success_timestamp_seconds gauge
ndt7_last_success_timestamp_seconds{client_ip="redacted",server="77.67.115.24:443"} 1.657630423e+09
# HELP ndt7_result_timestamp_seconds m-lab ndt7 test completion time in seconds since 1970-01-01
# TYPE ndt7_result_timestamp_seconds gauge
ndt7_result_timestamp_seconds{result="OK",test="download"} 1.657630413e+09
ndt7_result_timestamp_seconds{result="OK",test="upload"} 1.657630423e+09
# HELP ndt7_rtt_seconds m-lab ndt7 round-trip time in seconds
# TYPE ndt7_rtt_seconds gauge
ndt7_rtt_seconds 0.012
# HELP ndt7_upload_rate_bps m-lab ndt7 upload speed in bits/s
# TYPE ndt7_upload_rate_bps gauge
ndt7_upload_rate_bps 1.015494382638198e+08
```

I see the point of having a separate timestamp for the measurement's start time since it decouples the actual time of the measurement from the scraping time. However, I would put `client_ip` and `server` (hostname, rather than IP) on `ndt7_upload_rate_bps`, `ndt7_download_rate_bps` and `ndt7_rtt_seconds` instead, since they are properties of the measurement while `last_success_timestamp_seconds` looks like a property of the client itself. This allows grouping the results by server in PromQL very easily, and if a ~1m granularity (the default Prometheus scrape time) is enough, it won't require any join.

Also, why a separate `ndt7_last_success_timestamp_seconds` when there is `ndt7_result_timestamp_seconds` already?

In any case: this PR is getting large and complex enough that it would probably be a good idea to split it in two:

- one to factor out parts of the original main.go that will be reused in the prometheus exporter's main.go
- one to actually implement the prometheus exporter binary

What do you think?

> Reviewed 4 of 9 files at r1, 2 of 12 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-:-N555RQ98hQv_bryvi4-:bdtfdbu)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @fhltang)

___
*[`cmd/ndt7-client/ndt7-prometheus-exporter/main.go` line 190 at r3](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-N6wAXR9COt0srDThr-R:-N6wAXRA6A9skLS8QDsB:b-tsgs5q) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/1bb49d4d6a3a6e53fa267e40b55730618d992496/cmd/ndt7-client/ndt7-prometheus-exporter/main.go#L190)):*
> ```Go
> 		prometheus.MustRegister(lastResultGauge)
> 
> 		// The success gauge captures both the client IP and the server FQDN.
> ```

I think this would be better in README.md

fhltang:
> Hi, sorry about the delay, I've been quite busy lately.

No worries. I know how it can be like.


> You're right about the directory structure. Folders under `cmd/` should match the executable's name, so I believe what we want here is:
> 
> * cmd/ndt7-client/main.go
> * cmd/ndt7-prometheus-exporter/main.go
> * internal/...
> 
> There is an unofficial directory structure standard at https://github.com/golang-standards/project-layout#cmd which we try to follow. Regarding `internal`, it would probably be right to use the `internal/` folder at the root of the repository, which both `main.go` (the client's one and the exporter's one) can import and use.

Thanks for the reference. I think moving the shared code to the `internal/` directory at the root makes sense.

> If I'm not mistaken, these are all the metrics currently exported:
>
> ... _snipped_ ... 
> 
> I see the point of having a separate timestamp for the measurement's start time since it decouples the actual time of the measurement from the scraping time. However, I would put `client_ip` and `server` (hostname, rather than IP) on `ndt7_upload_rate_bps`, `ndt7_download_rate_bps` and `ndt7_rtt_seconds` instead, since they are properties of the measurement while `last_success_timestamp_seconds` looks like a property of the client itself. This allows grouping the results by server in PromQL very easily, and if a ~1m granularity (the default Prometheus scrape time) is enough, it won't require any join.

I had the same instinct as you. I think the best way to illustrate the problem with that approach is with this screenshot:

Different Client IP

This is a graph of a metric exported by the first version of this PR. By chance, my client IP changed (because I changed my broadband). Notice that the old test result stays around (until a process restart); this manifests as a very straight line in the graph.

The problem is that when a newer test result comes back, the code just records it as a gauge value but the previous test result hangs around. I think this is a consequence of the [`Gauge` interface](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#Gauge) I am using. I've seen other code use a Prometheus Collector API which might not have this issue but I'll need to research that.

My workaround is to make the result gauges have "last result" semantics and encode the client/server keys into a separate metric that can be joined.

> Also, why a separate `ndt7_last_success_timestamp_seconds` when there is `ndt7_result_timestamp_seconds` already?

Well spotted! (Good attention to detai.)

This annoyed me too when I was writing this code. This is an artefact of the `emitter.Emitter` interface. Metric `ndt7_result_timestamp_seconds` records both successes and failure events. `OnSummary` has the client and server details but only occurs on success, and `OnError` records failures but does not have client and server details. So there wasn't an easy to way to use a single metric that recorded success/failure and also client-server information.

Another option would have been to record the failure events in a separate metric. 
 
> In any case: this PR is getting large and complex enough that it would probably be a good idea to split it in two:
> 
> * one to factor out parts of the original main.go that will be reused in the prometheus exporter's main.go
> * one to actually implement the prometheus exporter binary
> 
> What do you think?

I think that makes sense. I would have made the same suggestion.

Let me work out how to split a PR into two. I will need to learn some more git voodoo. (I don't use git in my day job.)

While I work on that, I'd be interested to hear your thoughts on the metrics given my explanations above.

fhltang:
> The problem is that when a newer test result comes back, the code just records it as a gauge value but the previous test result hangs around. I think this is a consequence of the [`Gauge` interface](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#Gauge) I am using. I've seen other code use a Prometheus Collector API which might not have this issue but I'll need to research that.

Actually, I realise it would be trivial to "delete" the old labels. `prometheus.MetricVec` has a [`Reset()` method](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#MetricVec.Reset).

fhltang:
BTW, I hit https://github.com/golang/go/issues/27336 . This manifests as travis failing to generate the coverage report for coveralls because now there are two packages `cmd/ndt7-client` and `cmd/ndt7-prometheus-exporter` and they define the same flags, causing a collision.

I created commit 2a04224 as a workaround for lack of a better solution. It's not ideal because it now travis runs the tests in `cmd/` twice (and if you ever add another package under `cmd/`, you'd need to update `.travis.yml` to make sure you get coverage data).

fhltang (commented):
> Client and server IPs are subject to change and when this happens the exporter generates separate timeseries (like in the screenshot you posted), but you can always "group" them when querying. For example, you could sum by (instance, server) to get per-server timeseries from that instance of the exporter regardless of client_ip. In the situation above, you'd get a single time series, unless the server changes. If the server doesn't matter either, just sum by (instance) to get a single time series per exporter even if the server changes.

Unfortunately, the timestamp against each value in Prometheus is no longer the time of the test but rather the time when prometheus scraped the exporter server. At each scrape, prometheus fetches the test results from every client-server pair that exporter had tested *since its process started*. A consequence is I don't think there is a useful or meaningful [aggregation operator](https://prometheus.io/docs/prometheus/latest/querying/operators/#aggregation-operators) that can group these metrics:
- `sum()` doesn't make sense because that would add the test values together for different client-server pairs;
- `max()` at least picks one value but the value is the *greatest value* rather than most recent value;
- `avg()` aggregates meaningfully but we cannot control how many client-server pairs' test results we average; and
- the other aggregation operators have similar issues.

> In general, generating as many time series are needed to cover the expected use cases and then filtering/grouping in the PromQL query is the preferred way to go. In this case, some of the use cases I have in mind are:
> 
> - The user runs multiple measurement clients from different vantage points (N instances of this exporter)
>    - Each client has a different IP and, depending on the location, might hit a large set of different servers over time
> - The user wants a graph of the ul/dl bandwidth for each client_ip, regardless of the server (sum by instance, client_ip)
> - The user wants a graph of the ul/dl bandwidth for each instance and server regardless of client_ip changes (sum by instance, server)
> - The user wants to know the latest download speed for a specific client/server pair, or just for a specific M-Lab server

This is a great list of use cases. The use case that motivated me to build this server is the first one (except I only have one instance I actually care about).

I'm not sure if the exporter pattern built in this PR is the right solution for the last three use cases. The problem is that prometheus would only have values for client-server pairs that were present in the last scrape. Since this exporter is "stateless" (i.e. its state is not preserved between process restarts), the client-server pairs would be those that have been tested since the exporter server started.

For the last three use cases, you might be better served by a different exporter pattern, e.g. using Prometheus's Node Exporter to serve metrics from text files, see `--collector.textfile.directory` flag and https://github.com/prometheus/node_exporter/blob/master/collector/textfile.go . Using that  exporter pattern, you'd be using the filesystem as a persistence store. (Supposedly this is an abuse of Node Exporter's text file collector since Node Exporter is intended for node-level reporting.)

Another option would be to use the Prometheus Pushgateway. The 3rd bullet in section ["Should I be using the Pushgateway"](https://prometheus.io/docs/practices/pushing/#should-i-be-using-the-pushgateway) says series pushed to the Pushgateway are exposed forever. See also https://github.com/prometheus/pushgateway/blob/3dfc21fa8d9f5ae4cd48f86aed3f9568261284a7/README.md?plain=1#L214-L219

> Also, the time series are expected to stick around as long as the configured Prometheus retention time allows. Even if you Reset() the time series, it won't be exported anymore by this exporter but it'll stay in the Prometheus server since it was collected at some point in the past, and disappear only once the retention time has passed.

The time series still stick around in Prometheus but with historic timestamps. After an exporter process restarts, there won't be any data for client-server pairs for more recent timestamps.

I think you want to preserve state between an exporter process restart. See above for different exporter patterns.

> I would interpret a metric such as ndt7_download_rate_bps{client_ip="...",server="..."} as "the last recorded download speed between this client and this server", which would make sense to keep around.

If it helps, I can change the name of the metrics in this PR to clarify their meanings
- `ndt7_last_download_rate_bps`
- `ndt7_last_upload_rate_bps`
- `ndt7_last_rtt_seconds`

> > OnSummary has the client and server details but only occurs on success, and OnError records failures but does not have client and server details. So there wasn't an easy to way to use a single metric that recorded success/failure and also client-server information.
> 
> My first instinct here is that OnError should know more about the measurement that errored, or at least the information needed to update ndt7_result_timestamp_seconds. At the same time, this would also allow for better error messages in the other emitters, e.g. "Error during <download|upload> measurement to <server>: <error message>"

Yeah, I thought about that too. I'm wary of scope creep in this PR and chose to keep the `emitter` API as is.

> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-:-NAdpbq-9ei97p-PnVYf:b-g7h4we)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @fhltang)

fhltang (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/77)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @robertodauria)

___
*[`cmd/ndt7-client/ndt7-prometheus-exporter/main.go` line 190 at r3](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-N6wAXR9COt0srDThr-R:-NAeG7QO74tSmVBPgwJA:b-896fix) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/1bb49d4d6a3a6e53fa267e40b55730618d992496/cmd/ndt7-client/ndt7-prometheus-exporter/main.go#L190)):*

I think this would be better in README.md
</blockquote></details>

Done.

robertodauria (changes requested):
> Unfortunately, the timestamp against each value in Prometheus is no longer the time of the test but rather the time when prometheus scraped the exporter server. At each scrape, prometheus fetches the test results from every client-server pair that exporter had tested since its process started. A consequence is I don't think there is a useful or meaningful aggregation operator that can group these metrics: (snip)

You're right! I assumed you were deleting (i.e. `metric.Reset()`) the metric before updating it with the results from the current run. I've done a bit of experimentation, and if you reset the metric `sum by (instance)` works as I thought. It seems you had the right idea from the beginning by resetting the metric, and the queries can be kept very simple as long as you don't need the precise timestamp.

See below (I changed lastSuccess to be `ndt7_throughput_bps{client_ip, server, test}` for this experiment):

in the `OnSummary` method:
```
	p.lastSuccess.Reset()
	downloadSuccess := p.lastSuccess.WithLabelValues(
		s.ClientIP,
		fmt.Sprintf("%s:%s", s.ServerIP, s.ServerPort),
		"download",
	)
	downloadSuccess.Set(s.Download.Value * 1000.0 * 1000.0)
	uploadSuccess := p.lastSuccess.WithLabelValues(
		s.ClientIP,
		fmt.Sprintf("%s:%s", s.ServerIP, s.ServerPort),
		"upload",
	)
	uploadSuccess.Set(s.Upload.Value * 1000.0 * 1000.0)
``` 



Now, if I just want one continuous line with the throughput from this particular client:



Perhaps that one lower results was due to a different server?



...probably not, since the subsequent test (grey line) ran against the same server and resulted in a throughput more in line with the others.

Now, my IP changed. The `sum by (instance)` version still shows a continuous line:


(the zero value is due to the test running while I was resetting the network interface...)

but I can also check if the IP change made a difference:



With this approach, in the end the exported metrics would just be:
```
ndt7_throughput_bps{client_ip, server, test}
ndt7_rtt_seconds {client_ip, server, test}
ndt7_result_timestamp_seconds{result, test} 1.661791223e+09
```

Regarding exact timestamps VS Prometheus' collection time, I think your query which joins metrics from a few messages above would still work with `ndt7_throughput_bps` where getting the exact timestamp is important. However, please note we [ask users](https://www.measurementlab.net/develop/#best-practices-on-test-scheduling-and-frequency) to not run automated tests more than 4 times a day, while the platform actually starts limiting (interactive) tests after 40/day. Prometheus scrapes by default every 15s. Unless one sets  `scrape_interval` to some very high value, the delay probably won't make much of a difference to a user who just wants to monitor their connection automatically 4 times per day. A much simpler query is worth the small loss of precision in the general case, I think.

What do you think of the above? Anything I missed? Thanks!

> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-:-NAeKPB73TjcjvoFtej-:b1yolmz)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @fhltang and @robertodauria)

___
*[`internal/emitter/prometheus.go` line 80 at r7](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-NAeRbth6tLiRqB8payq:-NAeRbth6tLiRqB8payr:b-r8hm6n) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/b88795ba311be14f2f3fe5769adf1de9302b0314/internal/emitter/prometheus.go#L80)):*
> ```Go
> 	success := p.lastSuccess.WithLabelValues(
> 		s.ClientIP,
> 		fmt.Sprintf("%s:%s", s.ServerIP, s.ServerPort))
> ```

This code will require some changes to build after rebasing it on `main`,  i.e.:
- `s.Download` became `s.Download.Throughput.Value`
- `s.Download` could be nil to indicate that a download test hasn't been run -- same for the upload. In this case I think the exporter should just _not_ export the corresponding metric 
- The latency is per-test rather than just a single value (`s.Download.Latency.Value`, `s.Upload.Latency.Value`)
- Retransmission is only available for the download (`s.Download.Retransmission`)

fhltang (commented):
> See below (I changed lastSuccess to be ndt7_throughput_bps{client_ip, server, test} for this experiment):

That's a neat trick! Also, thanks for confirming that the `Reset()` call does the right thing.

> With this approach, in the end the exported metrics would just be:
> - `ndt7_throughput_bps{client_ip, server, test}`
> - `ndt7_rtt_seconds {client_ip, server, test}`
> - `ndt7_result_timestamp_seconds{result, test} 1.661791223e+09`

This makes sense. I have a slight hesitation about using `test` as a label to distinguish between download and upload subtests. The guidance in https://prometheus.io/docs/practices/naming/ states

> As a rule of thumb, either the sum() or the avg() over all dimensions of a given metric should be meaningful (though not necessarily useful).

Is the sum of download and upload throughput values meaningful? Maybe, e.g. if the network is "full duplex".

It's your call. I think it's important to get the metrics right because any changes in this aspect are likely to be incompatible.

> Regarding exact timestamps VS Prometheus' collection time, I think your query which joins metrics from a few messages above would still work with ndt7_throughput_bps where getting the exact timestamp is important.

I don't think the exact timestamp is important. The metric with completion times was just a workaround because I hadn't discovered the `Reset()` method. :)

> What do you think of the above? Anything I missed? Thanks!

I don't think you've missed.

Please confirm my question about the metrics and labels. I will merge the latest changes and adjust the metrics as necessary (in the evenings/weekends).

> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-:-NAfi8qg9UUF0MIRebb6:b-n0ohc)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @fhltang and @robertodauria)

robertodauria (commented):
> As a rule of thumb, either the sum() or the avg() over all dimensions of a given metric should be meaningful (though not necessarily useful).

TIL. :) Yeah, I don't think a sum/avg over `test` would be meaningful. So...
```
ndt7_download_throughput_bps{client_ip, server}
ndt7_upload_throughput_bps{client_ip, server}
ndt7_download_latency_seconds {client_ip, server}
ndt7_upload_latency_seconds {client_ip, server}

ndt7_result_timestamp_seconds{result, test} 1.661791223e+09
```

I'm not sure about `ndt7_result_timestamp_seconds`. Feel free to remove it if there's no use case for it anymore.

I've renamed `rtt` to `latency` because the latency in the summary is actually TCPInfo's `MinRTT` (minimum RTT recorded over the whole measurement) and just `rtt` would be a bit misleading.

> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-:-NAoLzcZDso7jIuM0e2R:bomqljt)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @fhltang and @robertodauria)

fhltang (commented):
Done. I've updated the metrics as discussed.

I also put some thought into the download retransmit metric is interesting. To export that to prometheus, I think we would need some more plumbing. At the moment, the emitter.SubtestSummary tuple stores [retransmission as a ratio](https://github.com/m-lab/ndt7-client-go/blob/f2d6e9baf0252da795a056330a02529318522107/internal/runner/runner.go#L160-L161) which cannot be aggregated directly (e.g. summing those ratios does not seem to be a meaninful way to aggregate). To export that in a way that can be aggregated, you would have to export the denominator as well, or just export the numerator (BytesRetrans) and denominator (BytesSent) which can be independently aggregated (using sum). I'll leave that to you if you are interested in that metric.

Another small bit of polish for `ndt7-prometheus-exporter` would be to add a handler for / which renders some recent test results in HTML. I can put that in a subsequent PR but I'll warn you I'm not good at HTML so the UI might hurt the eyes!

> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-:-N9r7yBv7SRSMnVcPGFi:b-83h9y6)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @robertodauria)

___
*[`internal/emitter/prometheus.go` line 80 at r7](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-NAeRbth6tLiRqB8payq:-NB4g47o2IZy6c3I0p0F:b-896fix) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/b88795ba311be14f2f3fe5769adf1de9302b0314/internal/emitter/prometheus.go#L80)):*

This code will require some changes to build after rebasing it on `main`,  i.e.:
- `s.Download` became `s.Download.Throughput.Value`
- `s.Download` could be nil to indicate that a download test hasn't been run -- same for the upload. In this case I think the exporter should just _not_ export the corresponding metric 
- The latency is per-test rather than just a single value (`s.Download.Latency.Value`, `s.Upload.Latency.Value`)
- Retransmission is only available for the download (`s.Download.Retransmission`)
</blockquote></details>

Done.

robertodauria (approved):
:lgtm: -- Thanks!

> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/77#-:-NCUp0u66EJ-R_1-zUpN:b-kejgc9)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @robertodauria)

PR #82, opened 2022-08-24
Show latency on upload-only measurements

robertodauria:
This PR changes the summary generation to display the latency from the server's TCPInfo after an upload-only test.

It also hides Retransmission in this case (since there is no way for the client to know it without accessing TCPInfo) and makes sure the client IP is always shown.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/82)

### DIFF
--- a/internal/emitter/humanreadable.go
+++ b/internal/emitter/humanreadable.go
@@ -91,43 +91,48 @@ func (h HumanReadable) OnComplete(test spec.TestKind) error {
 
 // OnSummary handles the summary event.
 func (h HumanReadable) OnSummary(s *Summary) error {
-	const summaryFormat = `%15s: %s
-%15s: %s
+	const summaryHeaderFormat = `
+Test results
+
+%10s: %s
+%10s: %s
+`
+	const downloadFormat = `
+%22s
+%15s: %7.1f %s
+%15s: %7.1f %s
 %15s: %7.1f %s
 `
-	_, err := fmt.Fprintf(h.out, summaryFormat,
+	const uploadFormat = `
+%20s
+%15s: %7.1f %s
+%15s: %7.1f %s
+`
+	_, err := fmt.Fprintf(h.out, summaryHeaderFormat,
 		"Server", s.ServerFQDN,
-		"Client", s.ClientIP,
-		"Latency", s.MinRTT.Value, s.MinRTT.Unit)
+		"Client", s.ClientIP)
 	if err != nil {
 		return err
 	}
 
-	if s.Download.Value != 0.0 {
-		_, err := fmt.Fprintf(h.out, "%15s: %7.1f %s\n",
-			"Download", s.Download.Value, s.Download.Unit)
+	if s.Download != nil {
+		_, err := fmt.Fprintf(h.out, downloadFormat, "Download",
+			"Throughput", s.Download.Throughput.Value, s.Download.Throughput.Unit,
+			"Latency", s.Download.Latency.Value, s.Download.Latency.Unit,
+			"Retransmission", s.Download.Retransmission.Value, s.Download.Retransmission.Unit)
 		if err != nil {
 			return err
 		}
 	}
 
-	if s.Upload.Value != 0.0 {
-		_, err := fmt.Fprintf(h.out, "%15s: %7.1f %s\n",
-			"Upload", s.Upload.Value, s.Upload.Unit)
+	if s.Upload != nil {
+		_, err := fmt.Fprintf(h.out, uploadFormat, "Upload",
+			"Throughput", s.Upload.Throughput.Value, s.Upload.Throughput.Unit,
+			"Latency", s.Upload.Latency.Value, s.Upload.Latency.Unit)
 		if err != nil {
 			return err
 		}
 	}
-
-	if s.DownloadRetrans.Value != 0.0 {
-		_, err := fmt.Fprintf(h.out, "%15s: %7.2f %s\n",
-			"Retransmission", s.DownloadRetrans.Value, s.DownloadRetrans.Unit)
-		if err != nil {
-			return err
-		}
-	}
-
-	fmt.Fprintln(h.out)
 
 	return nil
 }
--- a/internal/emitter/humanreadable.go
+++ b/internal/emitter/humanreadable.go
@@ -94,20 +94,40 @@ func (h HumanReadable) OnSummary(s *Summary) error {
 	const summaryFormat = `%15s: %s
 %15s: %s
 %15s: %7.1f %s
-%15s: %7.1f %s
-%15s: %7.1f %s
-%15s: %7.2f %s
 `
 	_, err := fmt.Fprintf(h.out, summaryFormat,
 		"Server", s.ServerFQDN,
 		"Client", s.ClientIP,
-		"Latency", s.MinRTT.Value, s.MinRTT.Unit,
-		"Download", s.Download.Value, s.Upload.Unit,
-		"Upload", s.Upload.Value, s.Upload.Unit,
-		"Retransmission", s.DownloadRetrans.Value, s.DownloadRetrans.Unit)
+		"Latency", s.MinRTT.Value, s.MinRTT.Unit)
 	if err != nil {
 		return err
 	}
+
+	if s.Download.Value != 0.0 {
+		_, err := fmt.Fprintf(h.out, "%15s: %7.1f %s\n",
+			"Download", s.Download.Value, s.Download.Unit)
+		if err != nil {
+			return err
+		}
+	}
+
+	if s.Upload.Value != 0.0 {
+		_, err := fmt.Fprintf(h.out, "%15s: %7.1f %s\n",
+			"Upload", s.Upload.Value, s.Upload.Unit)
+		if err != nil {
+			return err
+		}
+	}
+
+	if s.DownloadRetrans.Value != 0.0 {
+		_, err := fmt.Fprintf(h.out, "%15s: %7.2f %s\n",
+			"Retransmission", s.DownloadRetrans.Value, s.DownloadRetrans.Unit)
+		if err != nil {
+			return err
+		}
+	}
+
+	fmt.Fprintln(h.out)
 
 	return nil
 }
--- a/internal/emitter/humanreadable_test.go
+++ b/internal/emitter/humanreadable_test.go
@@ -2,7 +2,6 @@ package emitter
 
 import (
 	"errors"
-	"fmt"
 	"reflect"
 	"testing"
 
@@ -230,9 +229,6 @@ func TestHumanReadableOnSummary(t *testing.T) {
 	expected := `         Server: test
          Client: test
         Latency:    10.0 ms
-       Download:   100.0 Mbit/s
-         Upload:   100.0 Mbit/s
- Retransmission:    1.00 %
 `
 	summary := &Summary{
 		ClientIP:   "test",
@@ -261,12 +257,11 @@ func TestHumanReadableOnSummary(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	if len(sw.Data) != 1 {
-		t.Fatal("invalid length")
+	if len(sw.Data) == 0 {
+		t.Fatal("no data written")
 	}
+
 	if string(sw.Data[0]) != expected {
-		fmt.Println(string(sw.Data[0]))
-		fmt.Println(expected)
 		t.Fatal("OnSummary(): unexpected data")
 	}
 }
--- a/internal/emitter/humanreadable_test.go
+++ b/internal/emitter/humanreadable_test.go
@@ -226,28 +226,49 @@ func TestHumanReadableOnCompleteFailure(t *testing.T) {
 }
 
 func TestHumanReadableOnSummary(t *testing.T) {
-	expected := `         Server: test
-         Client: test
+	expectedHeader := `
+Test results
+
+    Server: test
+    Client: test
+`
+	expectedUpload := `
+              Upload
+     Throughput:   100.0 Mbit/s
         Latency:    10.0 ms
+`
+	expectedDownload := `
+              Download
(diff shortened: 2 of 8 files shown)


stephen-soltesz (commented):
> Reviewed 2 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/82)* status: 0 of 1 approvals obtained (waiting on @robertodauria)

___
*[`internal/emitter/humanreadable_test.go` line 231 at r2](https://reviewable.io/reviews/m-lab/ndt7-client-go/82#-NAFxEWfEUbQiVxv8k1q:-NAFxEWfEUbQiVxv8k1r:bohusni) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/ff06e896efe78d24b44d03414cc220c776149a4d/internal/emitter/humanreadable_test.go#L231)):*
> ```Go
> 	expected := `         Server: test
>          Client: test
>         Latency:    10.0 ms
> ```

Why does this not include the rates any more?

The new rate information should be in `sw.Data[1]`? 

I do not see other tests that check for the conditional formatting. Is this worth adding?
___
*[`internal/runner/runner.go` line 134 at r2](https://reviewable.io/reviews/m-lab/ndt7-client-go/82#-NAFy73_2yJ8qic6NJrp:-NAFy73_2yJ8qic6NJrq:b-7fqlll) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/ff06e896efe78d24b44d03414cc220c776149a4d/internal/runner/runner.go#L134)):*
> ```Go
> 
> 	var server, client string
> 	if results[spec.TestDownload].ConnectionInfo != nil {
> ```

You've removed the `!= nil` test. Is this guaranteed to be safe? (it may be, I'm asking to confirm)
___
*[`internal/runner/runner.go` line 151 at r2](https://reviewable.io/reviews/m-lab/ndt7-client-go/82#-NAFyjJH5jqiIRD3Rou6:-NAFyjJH5jqiIRD3Rou7:b1mxx0) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/ff06e896efe78d24b44d03414cc220c776149a4d/internal/runner/runner.go#L151)):*
> ```Go
> 
> 	if results[spec.TestDownload].ConnectionInfo != nil {
> 		s.DownloadUUID = results[spec.TestDownload].ConnectionInfo.UUID
> ```

Can this be done above also?

Also, it looks like we are processing counterflow messages now. But, this may not be the right PR to define UploadUUID though https://github.com/m-lab/ndt7-client-go/blob/main/internal/emitter/summary.go#L22
___
*[`internal/runner/runner.go` line 197 at r2](https://reviewable.io/reviews/m-lab/ndt7-client-go/82#-NAG-_lWDqMpKcywNOB9:-NAG-_lWDqMpKcywNOBA:b-qgpwp1) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/ff06e896efe78d24b44d03414cc220c776149a4d/internal/runner/runner.go#L197)):*
> ```Go
> 			if ul.Server.TCPInfo != nil {
> 				s.MinRTT = emitter.ValueUnitPair{
> 					Value: float64(ul.Server.TCPInfo.MinRTT) / 1000,
> ```

It's strange to me that this is conditional on the download test.

Could the Summary structure https://github.com/m-lab/ndt7-client-go/blob/main/internal/emitter/summary.go#L11  have directional summary stats so all stats could be returned for all combination of upload, download, download-then-upload tests?

robertodauria (commented):
@stephen-soltesz PTAL again?

> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/82#-:-NAKZAMj91tDsUVqfesi:b-hb6imj)* status: 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[`internal/emitter/humanreadable_test.go` line 231 at r2](https://reviewable.io/reviews/m-lab/ndt7-client-go/82#-NAFxEWfEUbQiVxv8k1q:-NAKXvE01df0JgFq33Gj:bnky99y) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/ff06e896efe78d24b44d03414cc220c776149a4d/internal/emitter/humanreadable_test.go#L231)):*

Why does this not include the rates any more?

The new rate information should be in `sw.Data[1]`? 

I do not see other tests that check for the conditional formatting. Is this worth adding?
</blockquote></details>

Probably not worth testing the format is exactly as expected, but I added it. I wouldn't be against removing it in subsequent PRs if it's too much of an inconvenience to keep it up-to-date.
___
*[`internal/runner/runner.go` line 134 at r2](https://reviewable.io/reviews/m-lab/ndt7-client-go/82#-NAFy73_2yJ8qic6NJrp:-NAKCw9LAF-ZXA0ABd7_:b-m2vepk) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/ff06e896efe78d24b44d03414cc220c776149a4d/internal/runner/runner.go#L134)):*

You've removed the `!= nil` test. Is this guaranteed to be safe? (it may be, I'm asking to confirm)
</blockquote></details>

It was since we were initializing both fields when creating the client instance. I've changed the client code so that those fields are only initialized when the corresponding test is run, so we can use their presence here to distinguish which tests were run. Checking that they aren't nil before using is now mandatory.
___
*[`internal/runner/runner.go` line 151 at r2](https://reviewable.io/reviews/m-lab/ndt7-client-go/82#-NAFyjJH5jqiIRD3Rou6:-NAKYVDP1HjyRKRVEc0I:bf6bqnr) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/ff06e896efe78d24b44d03414cc220c776149a4d/internal/runner/runner.go#L151)):*

Can this be done above also?

Also, it looks like we are processing counterflow messages now. But, this may not be the right PR to define UploadUUID though https://github.com/m-lab/ndt7-client-go/blob/main/internal/emitter/summary.go#L22
</blockquote></details>

I populate both UUID fields now (even if they're not displayed in the human-readable summary).
___
*[`internal/runner/runner.go` line 197 at r2](https://reviewable.io/reviews/m-lab/ndt7-client-go/82#-NAG-_lWDqMpKcywNOB9:-NAKYccA31PktSupofgZ:b92noh2) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/ff06e896efe78d24b44d03414cc220c776149a4d/internal/runner/runner.go#L197)):*

It's strange to me that this is conditional on the download test.

Could the Summary structure https://github.com/m-lab/ndt7-client-go/blob/main/internal/emitter/summary.go#L11  have directional summary stats so all stats could be returned for all combination of upload, download, download-then-upload tests?
</blockquote></details>

I've changed the summary structure quite a bit. Download/Upload substructures can be nil, so we don't use the zero value as a proxy for "this subtest has not been run" anymore (a retransmission of `0.0`, for example, is technically possible and not showing it at all in that case seemed weird).

stephen-soltesz (commented):
> Reviewed 3 of 6 files at r3.
> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/82)* status: 0 of 1 approvals obtained (waiting on @robertodauria)

___
*[`internal/emitter/humanreadable_test.go` line 231 at r2](https://reviewable.io/reviews/m-lab/ndt7-client-go/82#-NAFxEWfEUbQiVxv8k1q:-NAKxMHXBsIWYIsKXkji:b1o2aat) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/ff06e896efe78d24b44d03414cc220c776149a4d/internal/emitter/humanreadable_test.go#L231)):*

Probably not worth testing the format is exactly as expected, but I added it. I wouldn't be against removing it in subsequent PRs if it's too much of an inconvenience to keep it up-to-date.
</blockquote></details>

Thank you. I agree "exact" check isn't critical. Some sanity check that the information we expect is there is worthwhile I think. Thank you for adding it.
___
*[`internal/emitter/summary.go` line 9 at r3](https://reviewable.io/reviews/m-lab/ndt7-client-go/82#-NAKxdFZE5BABihjvevm:-NAKxdFZE5BABihjvevn:b-wurx68) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/79f1f79cc4b4785c2a0f54df436180047f09b262/internal/emitter/summary.go#L9)):*
> ```Go
> }
> 
> type SubtestSummary struct {
> ```

nit: docstring for exported type. This could include a note about all values being from the server perspective.
___
*[`internal/emitter/summary.go` line 13 at r3](https://reviewable.io/reviews/m-lab/ndt7-client-go/82#-NAK_ThK7HrAEsjRHDIY:-NAK_ThK7HrAEsjRHDIZ:b-bpeswe) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/79f1f79cc4b4785c2a0f54df436180047f09b262/internal/emitter/summary.go#L13)):*
> ```Go
> 	Throughput ValueUnitPair
> 	// Latency is the MinRTT value of the latest measurement, in milliseconds.
> 	// For uploads, this is provided by the server.
> ```

This is always provided by the server, correct?

Would a phrase like "Latency is always from the vantage point of the server, including for upload tests" be helpful?
___
*[`internal/emitter/summary.go` line 31 at r3](https://reviewable.io/reviews/m-lab/ndt7-client-go/82#-NAKxmuo-hWBdm8SeGbc:-NAKxmuo-hWBdm8SeGbd:byxugu5) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/79f1f79cc4b4785c2a0f54df436180047f09b262/internal/emitter/summary.go#L31)):*
> ```Go
> 	ClientIP string
> 
> 	Download *SubtestSummary
> ```

nit: simple comments here also would be welcome.
___
*[`internal/runner/runner.go` line 149 at r3](https://reviewable.io/reviews/m-lab/ndt7-client-go/82#-NAL-K_C-nXrmKnUpb71:-NAL-K_C-nXrmKnUpb72:bp6mx8f) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/79f1f79cc4b4785c2a0f54df436180047f09b262/internal/runner/runner.go#L149)):*
> ```Go
> 			appInfo := dl.Client.AppInfo
> 			elapsed := float64(appInfo.ElapsedTime) / 1e06
> 			s.Download.Throughput = emitter.ValueUnitPair{
> ```

just clarifying: download throughput comes from the client and everything else comes from the server?
___
*[`internal/runner/runner.go` line 204 at r3](https://reviewable.io/reviews/m-lab/ndt7-client-go/82#-NAKy1vPB4KCqwPleXdm:-NAKy1vPB4KCqwPleXdn:brvzhkh) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/79f1f79cc4b4785c2a0f54df436180047f09b262/internal/runner/runner.go#L204)):*
> ```Go
> 		s.ClientIP = clientIP
> 	}
> 	serverIP, _, err := net.SplitHostPort(server)
> ```

With CGNAT & server fail over, technically, client and server are not guaranteed to match for both upload and download cases.

If you'd like to address that in this PR, the FQDN/IPs could also be part of the `SubtestSummary`. If not, that's okay too.

robertodauria (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/82)* status: 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[`internal/emitter/summary.go` line 13 at r3](https://reviewable.io/reviews/m-lab/ndt7-client-go/82#-NAK_ThK7HrAEsjRHDIY:-NAdnpBT1eyzGI3z8LUZ:b-37jdar) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/79f1f79cc4b4785c2a0f54df436180047f09b262/internal/emitter/summary.go#L13)):*

This is always provided by the server, correct?

Would a phrase like "Latency is always from the vantage point of the server, including for upload tests" be helpful?
</blockquote></details>

I've added it to SubtestSummary's docstring. It may change in future, for now all the values except for the download throughput (measured at the receiver) are from the server.
___
*[`internal/runner/runner.go` line 149 at r3](https://reviewable.io/reviews/m-lab/ndt7-client-go/82#-NAL-K_C-nXrmKnUpb71:-NAdoNuU9EEe55iQPW1D:bx3yuy7) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/79f1f79cc4b4785c2a0f54df436180047f09b262/internal/runner/runner.go#L149)):*

just clarifying: download throughput comes from the client and everything else comes from the server?
</blockquote></details>

Correct. The throughput is always measured at the receiver as a rule.

Ideally, we would also have sender-side retransmission and latency during the upload, but the client must have access to the TCP_INFO struct for that -- and then it would not be multi-platform anymore, I suspect. As it is right now, this client behaves the same as the JS client (which will never be able to access TCP_INFO, running in a browser).
___
*[`internal/runner/runner.go` line 204 at r3](https://reviewable.io/reviews/m-lab/ndt7-client-go/82#-NAKy1vPB4KCqwPleXdm:-NAdpOsrCjGDiDvP7Les:bs1ttnz) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/79f1f79cc4b4785c2a0f54df436180047f09b262/internal/runner/runner.go#L204)):*

With CGNAT & server fail over, technically, client and server are not guaranteed to match for both upload and download cases.

If you'd like to address that in this PR, the FQDN/IPs could also be part of the `SubtestSummary`. If not, that's okay too.
</blockquote></details>

Ugh. I agree. I'd fix it in another PR, this one changes quite a lot already and did not introduce the bug (nor fix it, anyway :smile: )

stephen-soltesz (approved):
:lgtm:

> Reviewed 2 of 6 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/82#-:-NAdvroW4o7Lc-DQy3-h:bnfp4nl)* status: :shipit: complete! 1 of 1 approvals obtained

dtaht:
Thank you.

PR #89, opened 2023-03-09
Add source IP flag and implementation

maybe-hello-world:
This PR adds a "source" flag to the ndt7 client to allow running the test on a specific interface.


pvouzis:
Is this going to work on windows/mac/linux?

maybe-hello-world:
I tested it on linux only for our build, but I would expect it to work on mac and windows as well

PR #92, opened 2023-06-29
Send measurement if the test ended before UpdateInterval

cristinaleonr:
This PR adds functionality to send the measurement through the channel if one is received before `params.UpdateInterval`. It used to not be likely that a test would terminate before said interval, but this is now possible with the early-exit changes to the server.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/92)

### DIFF
--- a/internal/download/download.go
+++ b/internal/download/download.go
@@ -38,6 +38,12 @@ func Run(ctx context.Context, conn websocketx.Conn, ch chan<- spec.Measurement) 
 		}
 		mtype, r, err := conn.NextReader()
 		if err != nil {
+			elapsed := time.Now().Sub(start)
+			// If test finished before `params.UpdateInterval` and at least one message
+			// has been received, send its data before exiting.
+			if elapsed <= params.UpdateInterval && total > 0 {
+				sendMeasurement(ch, elapsed, total)
+			}
 			return err
 		}
 		var msgSize int64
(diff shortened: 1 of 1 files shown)


stephen-soltesz (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/92)* status: 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`internal/download/download.go` line 44 at r1](https://reviewable.io/reviews/m-lab/ndt7-client-go/92#-NZ7pSI94Wd8KNaIB6MV:-NZ7pSI94Wd8KNaIB6MW:byrj6b6) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/ccae5fae7327a3e7f535c2a466c86f9123871cde/internal/download/download.go#L44)):*
> ```Go
> 			// If the test finished before `params.UpdateInterval` and at least one message
> 			// has been received, send the measurement data through the channel before exiting.
> 			if elapsed <= params.UpdateInterval && total > 0 {
> ```

nit: there is a race here based on update interval and elapsed time. I can imagine scenarios where the elapsed time below is less than UpdateInterval but by the time we get here it is greater than.

What if we set a boolean below instead? And, report a measurement here only if that boolean is unset?

cristinaleonr (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/92)* status: 0 of 1 approvals obtained

___
*[`internal/download/download.go` line 44 at r1](https://reviewable.io/reviews/m-lab/ndt7-client-go/92#-NZ7pSI94Wd8KNaIB6MV:-NZamIgi7YfI17C8HwJu:b-896fix) ([raw file](https://github.com/m-lab/ndt7-client-go/blob/ccae5fae7327a3e7f535c2a466c86f9123871cde/internal/download/download.go#L44)):*

nit: there is a race here based on update interval and elapsed time. I can imagine scenarios where the elapsed time below is less than UpdateInterval but by the time we get here it is greater than.

What if we set a boolean below instead? And, report a measurement here only if that boolean is unset?
</blockquote></details>

Done.

stephen-soltesz (approved):
> *[Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/92)* status: :shipit: complete! 1 of 1 approvals obtained

PR #97, opened 2024-05-17
Service URLs may have a prefix

auvik-bheesham:
An example of this would be if the service is hosted behind some load balancer, or as a part of a larger service where there's a prefix .

This change makes the client a little more lenient, allowing for something like:

    go run ./cmd/ndt7-client/main.go -service-url 'wss://speedtest.example.com/some-prefix/ndt/v7/upload'

cc @pthankappan

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt7-client-go/97)

### DIFF
--- a/cmd/ndt7-client/main.go
+++ b/cmd/ndt7-client/main.go
@@ -184,10 +184,10 @@ func main() {
 	}
 
 	// If a service URL is given, then only one direction is possible.
-	if flagService.URL != nil && strings.Contains(flagService.URL.Path, params.DownloadURLPath) {
+	if flagService.URL != nil && strings.HasSuffix(flagService.URL.Path, params.DownloadURLPath) {
 		*flagUpload = false
 		*flagDownload = true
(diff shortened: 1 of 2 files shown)


cmd/ndt7-client/main.go
```diff
@@ -184,10 +184,10 @@ func main() {
 	}
 
 	// If a service URL is given, then only one direction is possible.
-	if flagService.URL != nil && strings.Contains(flagService.URL.Path, params.DownloadURLPath) {
+	if flagService.URL != nil && strings.HasSuffix(flagService.URL.Path, params.DownloadURLPath) {
```
auvik-bheesham, line 187: This is a change, I'm happy to revert this back to `Contains`, and sync `ndt7.go` along with it.

I figured it might be nice if the logic in both places was similar.

auvik-bheesham: Note that this also prevents something like:

```
ndt7-client -service-url 'wss://speedtest.example.com/ndt/v7/upload/ndt/v7/download`
```

